PR #18, opened 2026-04-03
Fix tv_launch for TradingView v2.14.0+ (Electron 38 / Node 22)

### LINKED ISSUE #13: tv_launch fails on TradingView Desktop v2.14.0 (Electron 38 / Node 22 strict flag validation)
## Summary

`tv_launch` and manual launch with `--remote-debugging-port=9222` fail on TradingView Desktop v2.14.0. The binary immediately exits with:

```
/Applications/TradingView.app/Contents/MacOS/TradingView: bad option: --remote-debugging-port=9222
```

## Root Cause

TradingView v2.14.0 ships with **Electron 38** which uses **Node.js v22.19.0**. Node 22 introduced [strict CLI flag validation](https://nodejs.org/en/blog/release/v22.0.0) that rejects unknown flags before Electron/Chromium can process them. `--remote-debugging-port` is a Chromium flag, not a Node flag, so Node rejects it at startup.

Running `--help` on the binary returns the Node.js help output (not Electron), confirming Node's flag parser runs first.

## What I Tried

| Approach | Result |
|----------|--------|
| Direct binary + flag | `bad option` exit |
| `open -a TradingView --args --remote-debugging-port=9222` | Launches but no CDP |
| `ELECTRON_EXTRA_LAUNCH_ARGS="--remote-debugging-port=9222"` (env var) | Launches but no CDP |
| `-- --remote-debugging-port=9222` (separator) | `Cannot find module '--remote-debugging-port=9222'` |
| `--inspect=9229` | Node debugger opens briefly, app crashes |

## Environment

- **macOS** Darwin 25.4.0 (Apple Silicon)
- **TradingView Desktop** v2.14.0 (Homebrew cask `tradingview`)
- **Electron** 38.2.2
- **Node** v22.19.0 (bundled with Electron)
- **tradingview-mcp** latest from main

## Possible Fixes

- Electron 38 may support `--remote-debugging-port` if passed via `app.commandLine.appendSwitch()` in the main process — but that requires modifying TradingView's `index.js` inside `app.asar` (which has integrity checks).
- Chromium supports reading flags from `chrome-flags.conf` — unclear if Electron 38 honors this.
- Older TradingView versions (pre-Electron 38) work fine with the flag. A version compatibility note in the README would help.
- There may be a Node 22 / Electron 38-specific env var or config file approach that bypasses strict validation.

This likely affects all users on TradingView v2.14.0+.

tradesdontlie:
## Summary

- **Detect spawn failure:** When `--remote-debugging-port` is rejected by Node 22's strict CLI validation (exit within 2s), automatically fall back to platform-specific strategies
- **macOS fallback:** Kill existing instance → `open -a TradingView.app --args --remote-debugging-port=9222` (confirmed working workaround via LaunchServices)
- **Linux/Windows fallback:** Re-spawn with `REMOTE_DEBUGGING_PORT` environment variable hint
- **Backward compatible:** Direct spawn tried first — old TradingView versions work exactly as before
- **Dependency injection:** Added `_deps` parameter to `launch()` for full testability without spawning real processes
- **20 new tests** covering every code path: path detection, kill logic, spawn success/failure, macOS `open -a` fallback, Linux/Win env var fallback, spawn error events, CDP polling, defaults

## Root Cause

TradingView v2.14.0 ships Electron 38 (Node 22), which rejects `--remote-debugging-port` as an unknown CLI flag before Chromium can process it. The key insight: `open -a` on macOS passes the flag through LaunchServices successfully, but **only if no existing instance is running** (otherwise macOS just activates the old non-CDP window).

## Test plan

- [x] All 49 tests pass (20 new + 29 existing)
- [x] Syntax check clean
- [x] No regressions in existing CLI, pine_analyze, or pine_check tests
- [ ] Manual verification on TradingView v2.14.0+ (macOS)
- [ ] Manual verification on older TradingView (backward compat)

Closes #13

### DIFF
--- a/src/core/health.js
+++ b/src/core/health.js
@@ -159,25 +159,26 @@ export async function uiState() {
   return { success: true, ...state };
 }
 
-export async function launch({ port, kill_existing } = {}) {
+export async function launch({ port, kill_existing, _deps } = {}) {
+  const deps = { spawn, execSync, existsSync, platform: process.platform, env: process.env, httpGet: null, ..._deps };
   const cdpPort = port || 9222;
   const killFirst = kill_existing !== false;
-  const platform = process.platform;
+  const platform = deps.platform;
 
   const pathMap = {
     darwin: [
       '/Applications/TradingView.app/Contents/MacOS/TradingView',
-      `${process.env.HOME}/Applications/TradingView.app/Contents/MacOS/TradingView`,
+      `${deps.env.HOME}/Applications/TradingView.app/Contents/MacOS/TradingView`,
     ],
     win32: [
-      `${process.env.LOCALAPPDATA}\\TradingView\\TradingView.exe`,
-      `${process.env.PROGRAMFILES}\\TradingView\\TradingView.exe`,
-      `${process.env['PROGRAMFILES(X86)']}\\TradingView\\TradingView.exe`,
+      `${deps.env.LOCALAPPDATA}\\TradingView\\TradingView.exe`,
+      `${deps.env.PROGRAMFILES}\\TradingView\\TradingView.exe`,
+      `${deps.env['PROGRAMFILES(X86)']}\\TradingView\\TradingView.exe`,
     ],
     linux: [
       '/opt/TradingView/tradingview',
       '/opt/TradingView/TradingView',
-      `${process.env.HOME}/.local/share/TradingView/TradingView`,
+      `${deps.env.HOME}/.local/share/TradingView/TradingView`,
       '/usr/bin/tradingview',
       '/snap/tradingview/current/tradingview',
     ],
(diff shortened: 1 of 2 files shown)


src/core/health.js
```diff
@@ -159,25 +159,26 @@ export async function uiState() {
   return { success: true, ...state };
 }
 
-export async function launch({ port, kill_existing } = {}) {
+export async function launch({ port, kill_existing, _deps } = {}) {
```
tansanDOTeth, line 162:
```suggestion
export async function launch({ port, kill_existing, deps } = {}) {
```

Might be a good idea to add a linter in here, but underscore is a convention for unused variable even though you are destructuring it in the next line.

55d9czt4sg-ui:
Great fix — the Electron 38 / Node 22 spawn detection and `open -a` fallback solves the root cause cleanly.

One small addition that pairs well with this: `tv_health_check` currently does a single-shot `TradingViewApi` access check and returns `api_available: false` if the chart JS isn't ready yet — which happens frequently right after a fresh launch. A retry loop makes it much more reliable:

```js
export async function healthCheck() {
  await getClient();
  const target = await getTargetInfo();

  // Retry TradingViewApi access — it may not be ready immediately after page load
  let state;
  for (let attempt = 0; attempt < 8; attempt++) {
    state = await evaluate(`
      (function() {
        var result = { url: window.location.href, title: document.title };
        try {
          var chart = window.TradingViewApi._activeChartWidgetWV.value();
          result.symbol = chart.symbol();
          result.resolution = chart.resolution();
          result.chartType = chart.chartType();
          result.apiAvailable = true;
        } catch(e) {
          result.symbol = 'unknown';
          result.resolution = 'unknown';
          result.chartType = null;
          result.apiAvailable = false;
          result.apiError = e.message;
        }
        return result;
      })()
    `);
    if (state?.apiAvailable) break;
    await new Promise(r => setTimeout(r, 1500));
  }

  return {
    success: true,
    cdp_connected: true,
    target_id: target.id,
    target_url: target.url,
    target_title: target.title,
    chart_symbol: state?.symbol || 'unknown',
    chart_resolution: state?.resolution || 'unknown',
    chart_type: state?.chartType ?? null,
    api_available: state?.apiAvailable ?? false,
  };
}
```

Up to 8 attempts × 1.5s = 12s max wait. In practice it resolves on the first or second attempt after a successful `tv_launch`. Without this, callers need their own retry logic or get a misleading `api_available: false` immediately after launch even when everything is fine.

PR #27, opened 2026-04-03
Improve Windows detection and runtime validation

### LINKED ISSUE #23: launch_tv_debug.bat doesn't detect Appx/WindowsApps installs
The JS-side `tv_launch` tool already uses `Get-AppxPackage` as a fallback for finding TradingView installed as an Appx package (health.js:200-209), but the bat script only tries `dir /s /b` on `WindowsApps`, which fails without elevated permissions.

### Expected behavior

`launch_tv_debug.bat` should find TradingView when installed as an Appx package under `C:\Program Files\WindowsApps\TradingView.Desktop_X.X.X.*\`.

### Suggested fix

Add a `Get-AppxPackage` PowerShell fallback between the `dir` and `where` checks, matching what `health.js` already does:

```bat
REM Use Get-AppxPackage for developer-sideloaded Appx installs (WindowsApps requires elevated dir)
if "%TV_EXE%"=="" (
    for /f "tokens=*" %%i in ('powershell -NoProfile -Command "Get-AppxPackage -Name '*TradingView*' | Select-Object -ExpandProperty InstallLocation" 2^>nul') do (
        if exist "%%i\TradingView.exe" set "TV_EXE=%%i\TradingView.exe"
    )
)
```

### Context

TradingView Desktop can be installed as a developer-signed Appx (`runFullTrust`) rather than from the MS Store. The versioned path (`TradingView.Desktop_3.0.0.7652_...`) lives under `WindowsApps` which requires admin to `dir`, but `Get-AppxPackage` works at user level.

### LINKED ISSUE #25: Enhancement: Validate against TradingView's live API state instead of hardcoded allowlists
## Problem

Several tools hardcode lists of valid values that TradingView actually exposes at runtime via internal APIs. When TradingView adds new options (chart types, layout codes, autoplay speeds), the hardcoded lists silently become stale — or worse, if TV removes a value, the tool sends an invalid value that corrupts cloud state.

### Current hardcoded values

| Tool | Hardcoded | Could query at runtime |
|------|-----------|----------------------|
| `replay_autoplay` | `[100, 143, 200, 300, 1000, 2000, 3000, 5000, 10000]` | `findAutoplayDelayOptionItems()` or similar internal |
| `chart_set_type` | Types 0–9 (`Bars`, `Candles`, ..., `HollowCandles`) | Chart type enum from TV internals |
| `pane_set_layout` | `s, 2h, 2v, 2-1, 1-2, 3h, 3v, 4, ...` | `_chartWidgetCollection` layout options |

## Best practice example

PR #24 (`replay_set_resolution`) demonstrates the runtime query pattern:

```js
// Query valid resolutions from TradingView — these are dynamic (depend on chart timeframe).
// Validate BEFORE calling changeReplayResolution to prevent cloud state corruption.
const available = await evaluate(wv(`${ctrl}._allReplayResolutions.value()`));
if (!Array.isArray(available))
  throw new Error('Could not retrieve available replay resolutions from TradingView.');
if (value !== null && !available.includes(value))
  throw new Error(`Invalid replay resolution "${interval}". Available: ${available.join(', ')}`);
```

This approach:
1. **Queries the live allowed list** from TradingView's own internal state
2. **Validates before mutating** — rejects bad input before any CDP call
3. **Returns helpful errors** — tells the user exactly what values are valid for their current context
4. **Never goes stale** — if TV adds or removes options, the validation updates automatically

## Proposal

Apply the same pattern to the hardcoded lists above:
- Query the actual enum/option set from TradingView's internal APIs at runtime
- Fall back to the current hardcoded list if the runtime query fails (defensive)
- Validate user input against the live list before any state-mutating CDP call

This eliminates an entire class of cloud state corruption bugs and keeps the tools in sync with TradingView updates automatically.

1cbyc:
## Summary

This PR bundles two fixes:

1. Windows launcher detection now handles Appx installs more reliably.
2. Validation for chart types, pane layouts, and replay autoplay speeds now prefers TradingView's live runtime state instead of trusting only hardcoded allowlists.

## What changed

- Added a `Get-AppxPackage` fallback to `scripts/launch_tv_debug.bat` so Appx/WindowsApps installs can be found without relying only on `dir` access.
- Updated the Windows detection note in the README.
- Updated `chart_set_type` to query TradingView runtime chart-type values when available, with fallback to the previous known-safe map.
- Updated `pane_set_layout` to query layout codes from TradingView runtime state when available, with fallback to the previous known-safe list.
- Updated `replay_autoplay` to query live autoplay delay options when available, with fallback to the previous known-safe delays.
- Kept the external tool/CLI surface stable while tightening validation before mutation.

## Verification

- `node --check src/core/chart.js`
- `node --check src/core/pane.js`
- `node --check src/core/replay.js`
- `node --check src/tools/chart.js`
- `node --check src/tools/pane.js`
- `node --check src/tools/replay.js`
- `npm run test:cli`

## Issues

Closes #23
Closes #25

### DIFF
--- a/README.md
+++ b/README.md
@@ -331,7 +331,7 @@ Launch scripts and `tv_launch` auto-detect TradingView. If auto-detection fails:
 | Platform | Common Locations |
 |----------|-----------------|
 | **Mac** | `/Applications/TradingView.app/Contents/MacOS/TradingView` |
-| **Windows** | `%LOCALAPPDATA%\TradingView\TradingView.exe`, `%PROGRAMFILES%\WindowsApps\TradingView*\TradingView.exe` |
+| **Windows** | `%LOCALAPPDATA%\TradingView\TradingView.exe`, `%PROGRAMFILES%\WindowsApps\TradingView*\TradingView.exe` (`Get-AppxPackage` fallback for Appx installs) |
 | **Linux** | `/opt/TradingView/tradingview`, `~/.local/share/TradingView/TradingView`, `/snap/tradingview/current/tradingview` |
 
 The key flag: `--remote-debugging-port=9222`
--- a/scripts/launch_tv_debug.bat
+++ b/scripts/launch_tv_debug.bat
@@ -21,13 +21,19 @@ REM Check MSIX / Windows Store installs
 if "%TV_EXE%"=="" (
     for /f "tokens=*" %%i in ('dir /s /b "%PROGRAMFILES%\WindowsApps\TradingView*\TradingView.exe" 2^>nul') do set "TV_EXE=%%i"
 )
+REM Use Get-AppxPackage for Appx installs when WindowsApps directory listing is blocked
+if "%TV_EXE%"=="" (
+    for /f "tokens=*" %%i in ('powershell -NoProfile -Command "Get-AppxPackage -Name '*TradingView*' | Select-Object -ExpandProperty InstallLocation" 2^>nul') do (
+        if exist "%%i\TradingView.exe" set "TV_EXE=%%i\TradingView.exe"
+    )
+)
 if "%TV_EXE%"=="" (
     for /f "tokens=*" %%i in ('where TradingView.exe 2^>nul') do set "TV_EXE=%%i"
 )
 
 if "%TV_EXE%"=="" (
     echo Error: TradingView not found.
-    echo Checked: %%LOCALAPPDATA%%\TradingView, %%PROGRAMFILES%%\TradingView, WindowsApps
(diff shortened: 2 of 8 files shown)


1cbyc:
The idea generally is to stop trusting stale hardcoded allowlists and instead validate against live runtime state. Like if we make runtime discovery the only path, then any minor TradingView internal rename breaks the tool completely, even when the old known-safe values would still work. That’s worse for users.

So the pattern I used is:
- Ask TradingView for the live allowed values.
- If TradingView exposes them, validate against that live list.
- If TradingView does not expose them, fall back to the previous knownsafe list instead of failing for no good reason.

That gives us both ease when TradingView exposes the data and strong resilience when TradingView internals move around. I hope this works for you guys? I already tested it. you can replicate.

tradesdontlie:
Thank you checking this out now

PR #20, opened 2026-04-03
Fix replay_stop and replay_autoplay corrupting account state

tradesdontlie:
## Summary

- **Remove all `hideReplayToolbar()` calls** from `stop()` and `start()` error recovery — this was syncing hidden-toolbar state to the user's TradingView cloud account, permanently breaking replay controls across all devices
- **Validate autoplay speed** against TradingView's 9 allowed values (`100, 143, 200, 300, 1000, 2000, 3000, 5000, 10000`) BEFORE any CDP call — invalid values corrupt `_autoplayDelay` in cloud state, causing assertion failures on every toolbar render
- **Fail fast** — validation moved before CDP connection so invalid input is rejected instantly with a clear error, no side effects

## Impact

Without this fix, affected users have **permanently broken replay controls** that survive uninstall/reinstall/cache clear (corruption lives in cloud-synced account state). The only recovery is manually patching `_autoplayDelay._value` via browser DevTools.

## Test plan

- [x] 23 new tests pass (9 valid delays accepted, 9 invalid delays rejected, 2 edge cases for speed=0/omitted, 3 source-level checks confirming hideReplayToolbar fully removed)
- [x] 29 existing tests pass (no regressions)
- [ ] Manual: `replay_stop` leaves toolbar visible
- [ ] Manual: `replay_autoplay` with `speed: 60000` returns clear validation error
- [ ] Manual: `replay_autoplay` with `speed: 1000` works normally

Closes #19

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #28, opened 2026-04-03
Fix replay state corruption safeguards

### LINKED ISSUE #19: CRITICAL: replay_stop and replay_autoplay corrupt TradingView account state, permanently breaking replay controls
## Bug

The `replay_stop` tool calls `hideReplayToolbar()` which syncs a hidden-toolbar state to the user's TradingView cloud account. Additionally, `replay_autoplay` accepts arbitrary delay values via `changeAutoplayDelay()`, but TradingView only supports 9 specific values: `100, 143, 200, 300, 1000, 2000, 3000, 5000, 10000`. Setting any other value (e.g. 60000) gets persisted to the cloud account and causes an assertion failure (`"This delay dosen't exist"`) every time the replay toolbar tries to render.

## Impact

- **Replay controls are permanently broken** across all devices and browsers — the corruption lives in the user's cloud-synced account state
- Full uninstall, reinstall, factory reset, and cache clearing do **not** fix it
- The only fix is manually patching the autoplay delay to a valid value via browser DevTools console
- TradingView support cannot easily diagnose this since the error is in minified client code

## Root cause

**`src/core/replay.js`**

1. **`stop()`** (lines 71, 75) — calls `hideReplayToolbar()` after stopping replay. TradingView persists toolbar visibility to the cloud account.

2. **`start()`** (line 37) — calls `hideReplayToolbar()` during error recovery.

3. **`autoplay()`** (line 58) — passes user-supplied `speed` directly to `changeAutoplayDelay()` with no validation. TradingView's internal `findAutoplayDelayOptionItems()` asserts if the value isn't in its hardcoded list.

## Steps to reproduce

1. Use `replay_start` then `replay_stop` — toolbar disappears
2. Use `replay_autoplay` with `speed: 60000` — replay controls crash on next open
3. Close and reopen TradingView — replay toolbar briefly flashes then collapses with assertion error
4. Same behavior in browser at tradingview.com — confirms cloud-synced corruption

## Suggested fix

1. Remove all `hideReplayToolbar()` calls — leave toolbar state to the user
2. Validate `speed` parameter against the valid delay values before calling `changeAutoplayDelay()`:

```js
const VALID_AUTOPLAY_DELAYS = [100, 143, 200, 300, 1000, 2000, 3000, 5000, 10000];
if (!VALID_AUTOPLAY_DELAYS.includes(speed))
  throw new Error(`Invalid autoplay delay. Valid values: ${VALID_AUTOPLAY_DELAYS.join(', ')}`);
```

## Recovery steps for affected users

Open browser DevTools console on `tradingview.com/chart` and run:

```js
var mgr = window.TradingViewApi._replayApi._replayUIController._replayManager;
mgr.changeAutoplayDelay(1000);
```

**Important:** Use `mgr.changeAutoplayDelay(1000)` (the proper API method), not `mgr._autoplayDelay._value = 1000` (direct mutation). Only the API method triggers the server-side save and propagates the fix account-wide across all layouts and devices.

1cbyc:
## Summary

This fixes the critical replay-state corruption bug described in #19.

## What changed

- Removed every `hideReplayToolbar()` call from `src/core/replay.js` so the tool no longer syncs a hidden replay toolbar state into the user's TradingView cloud account.
- Added strict pre-validation for `replay_autoplay` speeds using TradingView's known-safe delay list: `100, 143, 200, 300, 1000, 2000, 3000, 5000, 10000`.
- Added an offline regression test file covering both guardrails.

## Why

This code path can permanently poison replay controls across layouts, browsers, and devices because TradingView syncs that state through the account. Invalid autoplay delays are not a normal input error here; they are state corruption.

## Verification

- `node --check src/core/replay.js`
- `node --check tests/replay.test.js`
- `node --test tests/replay.test.js`

## Issues

Closes #19

### DIFF
--- a/src/core/replay.js
+++ b/src/core/replay.js
@@ -3,6 +3,8 @@
  */
 import { evaluate, getReplayApi } from '../connection.js';
 
+const VALID_AUTOPLAY_DELAYS = [100, 143, 200, 300, 1000, 2000, 3000, 5000, 10000];
+
 function wv(path) {
   return `(function(){ var v = ${path}; return (v && typeof v === 'object' && typeof v.value === 'function') ? v.value() : v; })()`;
 }
@@ -34,7 +36,6 @@ export async function start({ date } = {}) {
   if (toast) {
     // Stop replay to recover chart
     try { await evaluate(`${rp}.stopReplay()`); } catch {}
-    try { await evaluate(`${rp}.hideReplayToolbar()`); } catch {}
     throw new Error(`Replay date unavailable: "${toast}". The requested date has no data for this timeframe. Try a more recent date or switch to a higher timeframe (e.g., Daily).`);
   }
 
@@ -53,6 +54,10 @@ export async function step() {
 }
 
 export async function autoplay({ speed } = {}) {
+  if (speed > 0 && !VALID_AUTOPLAY_DELAYS.includes(speed)) {
+    throw new Error(`Invalid autoplay delay ${speed}. Valid values: ${VALID_AUTOPLAY_DELAYS.join(', ')}`);
+  }
+
   const rp = await getReplayApi();
   const started = await evaluate(wv(`${rp}.isReplayStarted()`));
   if (!started) throw new Error('Replay is not started. Use replay_start first.');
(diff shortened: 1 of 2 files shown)


1cbyc:
Like you didn't have to use try / await twice there to hide the replay tool bar. It just stresses the whole thing. I have tested this too. This is all good, you can review my PR, thank you @tradesdontlie

tradesdontlie:
Closing — the replay corruption fixes (hideReplayToolbar removal + autoplay delay validation) were merged in PR #20. Thank you for the contribution, your fix was identical to ours!

PR #35, opened 2026-04-04
feat: add data_get_pine_shapes for reading plotshape/plotchar signals

tansanDOTeth:
New tool that reads plotshape() and plotchar() markers from Pine Script indicators. Unlike box.new/label.new/line.new which store data in _primitivesCollection, plotshape outputs are stored in the study bar data series. This tool scans the study metaInfo for shapes type plots, then reads the bar data to find which bars have active markers.

Returns per signal: plot title, shape type, location, color, bar index, and OHLC of the candle the signal fired on. Configurable scan depth via last_n_bars param (default 100).

Also returns metadata about all shape plots defined by the indicator so the LLM knows what signals the indicator can produce.


tradesdontlie:
amazing checking this out now thank you

PR #44, opened 2026-04-06
feat: ASCII price action chart tools to improve LLM market structure understanding

BurnOutTrader:
## Summary

- **Port signal bar detection to match proper breakout conditions** — the previous implementation flagged nearly every higher-high bull bar as a signal, producing noise on every bar. The new implementation gates signals on six conditions: close breaking through the prior bar's high/low, a strength threshold (close within 16% of range from the extreme), a minimum consecutive run of opposing bars in a lookback window, a recent opposing bar gate, and an interleaving guard that prevents back-to-back same-direction signals. This results in sparse, meaningful signals that reflect genuine momentum shifts.
- **Fix `buildBarAnatomyASCII` connector alignment** — the `──` connectors and vertical continuation lines previously used hardcoded spacing, causing misalignment when price values had different string lengths. Now uses dynamic padding computed from the actual price string width so all connectors appear at a consistent column.
- **Round Y-axis tick labels** — chart Y-axis labels now snap to round intervals (e.g. `4730 / 4720 / 4710`) instead of raw floating-point values, making the price scale easier to read.
- **Dedicated swing marker rows** — H/L swing markers are now rendered in separate rows above and below the chart rather than inline with candle bodies, eliminating visual collision with price action characters.
- **Update README examples with live 1H data** — all chart examples replaced with real output.

## Why ASCII price action tools help the LLM

The MCP already provides `capture_screenshot` for visual context and data tools (`data_get_ohlcv`, `data_get_study_values`, etc.). The chart analysis tools in this PR address a specific gap: **giving the model structured, reasoned price action data it can work with directly in text, without relying on vision**.

A screenshot requires the model to visually interpret candle positions, wick ratios, and pattern placement — and vision inference is lossy. The ASCII chart tools provide the same information as structured, unambiguous text:

- Every bar's direction, HH/HL/LH/LL label relative to the prior bar, candlestick pattern classification, and volume character
- Confirmed swing points with their structural label in the sequence
- Trend lines with exact slope and projected current price
- Signal bar positions with count — so the model knows whether the market is on a first attempt or a confirmed follow-through
- Volume profile with POC, Value Area, and buying vs selling split at each price level

This means the model can reason about market structure with the same precision it applies to code or data — not approximating from pixels, but reading exact values. Combined with a screenshot for gestalt context, the two approaches are complementary: the screenshot gives the visual picture, the ASCII tools give the numbers the model needs to generate accurate, specific analysis.

**`chart_structure` + `capture_screenshot` workflow:**
1. `capture_screenshot` → model sees the visual layout and overall shape
2. `chart_structure` → model gets exact swing sequence, trend line slopes, S/R levels, signal bar positions
3. `chart_volume_profile` → model anchors visual S/R levels to actual volume acceptance zones
4. `chart_individual_bar` → model quantifies wick/body ratios and pattern classification for the bar it's discussing

Without the ASCII tools, the model is guessing at structure. With them, it can give specific, defensible price action analysis grounded in the actual data.

## Test plan

- [ ] Restart MCP server and verify `chart_structure` signal rows are sparse across a 60-bar window
- [ ] Confirm `chart_individual_bar` ASCII diagram has all `──` connectors at the same column across varying price string lengths
- [ ] Confirm Y-axis labels show round numbers on both `chart_price_action` and `chart_structure`
- [ ] Run `capture_screenshot` then `chart_structure` on the same chart and verify the model's analysis references both the visual and the structured data

### DIFF
--- a/README.md
+++ b/README.md
@@ -68,6 +68,264 @@ Gives your AI assistant eyes and hands on your own chart:
 - **CLI access** — every MCP tool is also a `tv` CLI command, pipe-friendly with JSON output
 - **Launch TradingView** — auto-detect and launch with debug mode from any platform
 
+## AI Market Analysis
+
+In raw images the ai will mistake prices.LLM's are better at recognising patterns in ASCII than in raw data or images.
+Four tools give Claude a structured, text-based view of price action that it can reason over directly — no screenshot required. All tools require a candlestick-based chart type (Candles, Bars, Hollow Candles, Heikin Ashi). Renko, Kagi, P&F, and Line Break charts are unsupported. 
+
+Combined with chart images it helps the LLM to accurately determine prices and price action sequences and patterns.
+
+---
+
+### `chart_structure` — Swing Structure, Trend Lines, S/R, Al Brooks Signals
+
+The primary analysis tool. Detects swing highs and lows using a ±lookback bar window, labels each confirmed swing as **HH / LH / HL / LL** relative to the prior swing, derives trend lines through consecutive swing points, extracts support and resistance levels, and overlays **Al Brooks consecutive H/L signal bar counts** (H1/H2/H3/H4 = consecutive bullish signal bars; L1/L2/L3/L4 = consecutive bearish).
+
+```
+   4825.70 ┤                                     H         H │
+           ┤                                 █ █ ▓       │ █ █ ▓
+           ┤                               │ █ │ ▓ ▓ │ │ █ █ │ ▓
+           ┤                 H       █ │   █ █     ▓ █ ▓ █     ▓
+   4774.04 ┤                 │ │     █ ▓ │ █       ▓ █ │ │     ▓
+           ┤               █ ▓ █ ▓ █ █ ▓ █ █       L           ▓
+           ┤   │ │         █ ▓ █ ▓ █   ▓ █ │                   ▓                                                             H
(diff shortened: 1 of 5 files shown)


tansanDOTeth:
I had a similar issue to you, but I took a different route. I extended some of the MCPs ability to detect the shapes or drawings that I needed from my pinescripts in #35, so then the MCP can query the correct metadata I needed to figure out some numbers.

One query that I had was, "could you measure from the recent blue square to the current price?" 

I imagine you can probably extend the MCP to parse out drawings/shapes/lines/etc... with its associated candle meta data so help you achieve this

BurnOutTrader:
It is probably possible, I have been using ASCII in my automated strategies for a long time. 
I have a lot more scripts in python each script has a specific prompt.
When you combine them it is: 
1. Teaching the model how to read what you are showing it.
2. Showing price as a structured image the model can read with more price accuracy. I have had issues in the past with incorrect prices from pure images alone. 
3. LLMs can whip up scripts in the background to do quant work from raw data, but it is token heavy and they perform worse than just using simple ML models. They are not great at determining probabilities from raw price data.

I just had claude move a few samples of my code from python scripts into the code base, i am not sure if you will find it helpful or not. The reports are fun. I have a similar MCP hooked up to my automated engine to monitor strategies and accounts while i sleep.

PR #49, opened 2026-04-07
Fix getChartApi not defined in drawing management functions

andreist257:
## Summary

- `draw_list`, `draw_clear`, `draw_remove_one`, and `draw_get_properties` all fail with `getChartApi is not defined`
- Root cause: these four functions call `getChartApi()` and `evaluate()` directly, but both are imported as `_getChartApi` and `_evaluate` — only `drawShape()` properly calls `_resolve(_deps)` to bring them into scope
- Fix: add `_deps` parameter and `_resolve(_deps)` call to all four functions, matching the existing `drawShape()` pattern

## Test plan

- [x] `draw_list` — returns all shapes on chart
- [x] `draw_remove_one` — removes shape by entity ID
- [x] `draw_clear` — removes all shapes
- [x] `draw_get_properties` — returns shape properties

All four were manually tested via MCP after the fix.

### DIFF
--- a/src/core/drawing.js
+++ b/src/core/drawing.js
@@ -44,7 +44,8 @@ export async function drawShape({ shape, point, point2, overrides: overridesRaw,
   return { success: true, shape, entity_id: result?.entity_id };
 }
 
-export async function listDrawings() {
+export async function listDrawings({ _deps } = {}) {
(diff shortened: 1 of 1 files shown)


tradesdontlie:

Issue #48, opened 2026-04-07
Strategy tester API endpoints are broken in TradingView Desktop 2.14.0


bangkokskater:
Strategy tester API endpoints are broken in TradingView Desktop 2.14.0. This is confirmed in the latest version. The old internal strategy tester API paths were removed and are not available anymore. 
You need to downgrade TradingView Desktop to version 2.13.1 for strategy reading tools to work. All other chart tools work correctly.

PR #51, opened 2026-04-08
feat: improve strategy detection and add DOM metrics fallback

hoonstake:
## Summary

`getStrategyResults()` was failing to find metrics for many strategies. Root causes:

1. **Overlay strategies missed**: Strategies with `overlay=true` have `is_price_study=true`, so the old `is_price_study===false` filter skipped them entirely
2. **Volume false positive**: The Volume indicator has `is_price_study=false` and was matched before the actual strategy
3. **Reactive wrapper issues**: `reportData()` returns an object with `.performance.all` containing the actual metrics, but the old code tried shallow key extraction that missed nested data

### `src/core/data.js`

**`getStrategyResults()` rewrite:**
- Uses `ordersData` as primary strategy detection (only strategies have this property)
- Falls back to `scriptType` and `is_price_study` with built-in study filtering
- Checks strategy state (`isFailed()`, `isLoading()`) before extraction
- Extracts from `reportData().performance.all` which contains all 30+ metrics (netProfit, totalTrades, profitFactor, etc.)
- Returns debug info when extraction fails to aid troubleshooting

**New: `getStrategyMetricsFromDom()`:**
- DOM-based fallback that scrapes the Strategy Tester panel
- Auto-clicks the metrics tab ("지표" / "Performance Summary")
- Supports English and Korean labels
- Returns same shape as `getStrategyResults()`

**New: `evaluateJs()`:**
- Generic JS evaluation via CDP for debugging

### `src/cli/commands/data.js`
- Added `strategy-dom` subcommand
- Added `evaluate-js` subcommand

### `src/tools/data.js`
- Registered `data_get_strategy_metrics_dom` MCP tool
- Registered `data_evaluate_js` MCP tool

## Testing

Tested with:
- `overlay=true` strategy (EMA cross) — was broken, now returns 44 metrics
- `overlay=false` strategy — still works
- Chart with no strategy — returns clear error message
- Failed/broken strategy — returns state info instead of empty metrics
- DOM fallback — reads metrics when API path fails
- Volume indicator on chart — correctly skipped

## Backward Compatibility

- `getStrategyResults()` returns the same shape: `{success, metric_count, source, metrics, error}`
- Optional `debug` field added only when metrics are empty
- Optional `state` field added only for failed/loading strategies

### DIFF
--- a/src/cli/commands/data.js
+++ b/src/cli/commands/data.js
@@ -24,7 +24,7 @@ register('values', {
 });
 
 register('data', {
-  description: 'Advanced data tools (lines, labels, tables, boxes, strategy, trades, equity, depth)',
+  description: 'Advanced data tools (lines, labels, tables, boxes, strategy, trades, equity, depth, evaluate-js, strategy-dom)',
   subcommands: new Map([
     ['lines', {
       description: 'Get Pine Script line.new() price levels',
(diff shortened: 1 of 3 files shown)


tradesdontlie:

PR #59, opened 2026-04-12
feat: add switchScript for UI-based Pine editor context switching

prezis:
## Summary

- Adds `switchScript(name)` to `pine.js` — switches the Pine editor UI to a different script using the dropdown menu
- Unlike `openScript()` which only fetches code via API, this actually switches the UI editor context
- Critical for multi-indicator workflows where you need to edit/deploy different scripts without overwriting the wrong one
- Includes `getActiveScriptName()` helper that reads the current script name from the nameButton

## How it works

1. Clicks the script nameButton to open dropdown
2. Finds the target script by text matching
3. Clicks its coordinates via CDP
4. Verifies the switch by re-reading nameButton text
5. Short-circuits if already on target script

## Test plan

- [x] Tested with 3 simultaneous Pine indicators
- [x] Verified correct script switching via dropdown
- [x] Verified no overwrite of wrong script after switch
- [x] Handles "already on target" case (no-op)

### DIFF
--- a/src/core/pine.js
+++ b/src/core/pine.js
@@ -588,6 +588,75 @@ export async function openScript({ name }) {
   return { success: true, name: result.name, script_id: result.id, lines: result.lines, source: 'internal_api', opened: true };
 }
 
+/**
+ * Switch Pine editor to a different saved script via UI dropdown.
+ * This properly switches the editor context (unlike openScript which just sets code).
+ *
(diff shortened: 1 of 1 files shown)


prezis:
Closing — no reviews after 10+ days. The switchScript feature works in our fork. Happy to reopen if there's interest.

PR #62, opened 2026-04-13
fix(drawing): restore DI in listDrawings, getProperties, removeOne, clearAll (regression from f23eb1b)

lrdplopes:
`draw_list`, `draw_get_properties`, `draw_remove_one`, and `draw_clear` throw `ReferenceError: getChartApi is not defined` at runtime.

It seems a partial DI refactor in `f23eb1b` aliased imports in `src/core/drawing.js`, but `_resolve(_deps)` was only wired into `drawShape`. This patch applies the same pattern to the remaining four functions and adds regression tests.


tradesdontlie:

PR #65, opened 2026-04-14
feat: add watchlist_remove, watchlist_add_bulk, fix click handling

drm-collab:
## Summary
- **`watchlist_remove`**: Remove one or more symbols by name. Handles exchange-prefixed tickers (`NYSE:AAPL` → matches `AAPL`). Cleans up stale search overlays on error.
- **`watchlist_add_bulk`**: Add multiple symbols in one batch call with configurable delay between additions and per-symbol success/failure reporting.
- **Click handling fix**: Replaces all `.click()` calls with proper `mousedown → mouseup → click` MouseEvent dispatches. TradingView Desktop (Electron 38+) ignores synthetic `.click()` because React event handlers require real mouse events with coordinates.

## Motivation
The existing `watchlist_add` tool works for single symbols but is tedious when populating a watchlist from a screener (20+ symbols). `watchlist_add_bulk` handles the batch case. `watchlist_remove` was missing entirely. The click fix resolves a regression where watchlist operations silently failed on recent Electron/TV Desktop versions.

## Test plan
- [ ] `watchlist_add` works on Electron 38+ (real mouse events)
- [ ] `watchlist_add_bulk` adds 5+ symbols with per-symbol status
- [ ] `watchlist_remove` removes a symbol by bare ticker
- [ ] `watchlist_remove` removes exchange-prefixed symbol
- [ ] CLI: `tv watchlist add-bulk AAPL MSFT GOOGL` works
- [ ] CLI: `tv watchlist remove AAPL` works
- [ ] Error recovery: stale search overlay dismissed on failure

### DIFF
--- a/src/cli/commands/watchlist.js
+++ b/src/cli/commands/watchlist.js
@@ -2,7 +2,7 @@ import { register } from '../router.js';
 import * as core from '../../core/watchlist.js';
 
 register('watchlist', {
-  description: 'Watchlist tools (get, add)',
+  description: 'Watchlist tools (get, add, add-bulk)',
   subcommands: new Map([
     ['get', {
       description: 'Get watchlist symbols',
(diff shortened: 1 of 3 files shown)


tradesdontlie:
Thanks @drm-collab — your React-fiber watchlist probe and REST-API removal design were exactly right, and they're now on main (with credit) via #327. The final version also handles two things discovered during live verification: TV renamed the right-rail button (data-name is now "base"), and the desktop widget doesn't live-sync API removals, so the panel is remounted and the removal verified. watchlist_remove and watchlist_add_bulk are both live. Closing as superseded.

PR #73, opened 2026-04-16
Auto-detect TradingView when installed as MSIX (Microsoft Store)

Rooney1419:
## Summary

`launch()` and `tv_launch` on Windows miss TradingView when it's installed from the Microsoft Store. Fix by asking the AppX system for the InstallLocation and wiring that into the candidate list.

## The bug

```
$ tv launch
{ "success": false,
  "error": "TradingView not found on win32. Searched: C:\Users\...\TradingView\TradingView.exe, C:\Program Files\TradingView\TradingView.exe, C:\Program Files (x86)\TradingView\TradingView.exe." }
```

…even though TradingView is installed and runnable. Microsoft Store installs live at:

```
C:\Program Files\WindowsApps\TradingView.Desktop_<version>_x64__<pubhash>\
```

…which isn't in the path scan. The fallback to `where TradingView.exe` also doesn't help because the MSIX package isn't on PATH.

Can't just add `WindowsApps` to the candidates with a glob because **the folder is ACL-protected** — `readdirSync` throws `EPERM` for non-admin users, so you can't enumerate it from Node.

## The fix

Ask the AppX system via `Get-AppxPackage`, which always works for the current user's installed packages regardless of ACLs:

```js
function findMsixTradingView() {
  try {
    const out = execSync(
      'powershell -NoProfile -NonInteractive -Command ' +
      '"(Get-AppxPackage -Name TradingView.Desktop | Select-Object -First 1).InstallLocation"',
      { timeout: 5000, stdio: ['ignore', 'pipe', 'ignore'] }
    ).toString().trim();
    if (!out) return null;
    const exe = join(out, 'TradingView.exe');
    return existsSync(exe) ? exe : null;
  } catch {
    return null;
  }
}
```

And append its return value to the `win32` candidate array. Existing code already handles `null` entries via `if (p && existsSync(p))`, so:

- No MSIX install → helper returns `null` → no-op
- Non-Windows hosts → function is never called (the `pathMap[platform]` branch doesn't reach it)
- PowerShell unavailable → `execSync` throws → caught → `null` → no-op

## Test plan

- [x] Verified on Windows 11 with Store install `TradingView.Desktop_3.0.0.7652_x64__n534cwy3pjxzj` — helper returns the correct exe path
- [x] `npm run test:unit` — 29/29 pass
- [x] `npm run test:cli` — 13/13 pass
- [x] No new dependencies (uses existing `execSync`, `existsSync`, adds only `path.join`)

## Also applies to

This MSIX path scan problem equally affects macOS Mac App Store installs of Electron apps, but that's a separate concern — TradingView Desktop on Mac is distributed as a `.app` bundle not through the App Store, so no action needed there.

### DIFF
--- a/src/core/health.js
+++ b/src/core/health.js
@@ -3,7 +3,30 @@
  */
 import { getClient, getTargetInfo, evaluate } from '../connection.js';
 import { existsSync } from 'fs';
+import { join } from 'path';
 import { execSync, spawn } from 'child_process';
+
+/**
+ * Find TradingView.exe inside the Windows MSIX (Microsoft Store) install
+ * location. The WindowsApps folder is ACL-protected so readdir throws EPERM
+ * for non-admin users — ask the AppX system instead via Get-AppxPackage,
+ * which always works for the current user's installed packages.
+ * Returns the absolute path to TradingView.exe or null.
+ */
+function findMsixTradingView() {
+  try {
+    const out = execSync(
+      'powershell -NoProfile -NonInteractive -Command ' +
+      '"(Get-AppxPackage -Name TradingView.Desktop | Select-Object -First 1).InstallLocation"',
+      { timeout: 5000, stdio: ['ignore', 'pipe', 'ignore'] }
+    ).toString().trim();
+    if (!out) return null;
+    const exe = join(out, 'TradingView.exe');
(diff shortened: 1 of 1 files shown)


tradesdontlie:

PR #76, opened 2026-04-18
fix(windows): support MSIX-packaged TradingView Desktop in tv_launch

### LINKED ISSUE #14: Windows App
Hi, it fails to run if installed as msix, and I don't think we have exe standalone of TradingView?
my path is like this: C:\Program Files\WindowsApps\TradingView.Desktop_3.0.0.7652_x64__n875hsr8pjxzj\TradingView.exe

👉 The Microsoft Store (MSIX) build of TradingView will not expose CDP, no matter what flags you pass.

### LINKED ISSUE #23: launch_tv_debug.bat doesn't detect Appx/WindowsApps installs

### LINKED ISSUE #42: Windows MSIX: CDP fails with --remote-debugging-port (access denied / 0x800704C7)
## Problem

TradingView Desktop on Windows is now exclusively MSIX (Microsoft Store). 
The download from tradingview.com/desktop installs to WindowsApps — 
no classic installer exists anymore.

Installed path:
C:\Program Files\WindowsApps\TradingView.Desktop_3.0.0.7652_x64__n534cwy3pjxzj\TradingView.exe

## Everything tried

1. Direct execution → Access Denied (even as admin)
2. launch_tv_debug.bat → same
3. Invoke-CommandInDesktopPackage with correct PackageFamilyName + AppId:

Invoke-CommandInDesktopPackage `
  -PackageFamilyName "TradingView.Desktop_n534cwy3pjxzj" `
  -AppId "TradingView.Desktop" `
  -Command "TradingView.exe" `
  -Args "--remote-debugging-port=9222"

→ COMException 0x800704C7 (ERROR_CANCELLED)
  Even with Developer Mode ON and running as Administrator.

## Questions

- Is there a pre-MSIX installer (v2.x) that used the classic Squirrel format?
- Does this tool currently work on Windows at all, or only Mac/Linux?
- Any alternative to the flag? (registry key, user-data dir config, env var?)

## Environment
Windows 10 Pro 10.0.19045 · TradingView Desktop 3.0.0.7652 (MSIX) · Developer Mode: enabled

unknowntrader7:
## Summary

Adds Windows MSIX support to `tv_launch` so it works on the current TradingView Desktop distribution without external scripts.

TradingView Desktop on Windows now ships **exclusively as an MSIX package** — even when downloaded from tradingview.com/desktop/ (not Microsoft Store). MSIX apps install under `C:\Program Files\WindowsApps\` where Windows blocks direct `.exe` invocation with command-line arguments for security reasons. This means the existing logic in `launch()`:

```js
spawn(tvPath, [`--remote-debugging-port=${cdpPort}`], { detached: true, ... })
```

silently fails on every current Windows install — the `.exe` path isn't even discoverable since the scanned paths (`%LOCALAPPDATA%\TradingView`, `%PROGRAMFILES%\TradingView`) no longer exist. Users see `tv_health_check` → *CDP fetch failed* and MCP servers marked disconnected.

## The fix

Use Microsoft's official `IApplicationActivationManager` COM API, which is the sanctioned way to activate packaged apps with arguments. The flag goes through the sandbox cleanly.

- **`findMsixTradingView()`** — detects the TradingView.Desktop MSIX package via `Get-AppxPackage` and returns the package family name + AUMID.
- **`launchMsixTradingView()`** — invokes a parametrized PowerShell helper (`scripts/launch_msix.ps1`) that activates via COM and returns JSON with the process ID.
- **`launch()`** — tries legacy native paths first (no change for users on older installs), falls back to MSIX only when nothing is found. Return object gains `launch_method: 'native' | 'msix'` for observability.
- **e2e test** — the `tv_launch — auto-detect binary` test previously only looked at hardcoded Mac paths and failed on any Windows machine. Now platform-aware and accepts MSIX package presence as a valid discovery result.
- **README** — adds an MSIX note to the Windows quick-start and the install-location table. Keeps the existing `launch_tv_debug.bat` docs but flags that it won't find MSIX installs and points users at `tv_launch` instead.

## Verified locally

Windows 11 Home, TradingView Desktop 3.0.0 MSIX (PackageFamilyName `TradingView.Desktop_n534cwy3pjxzj`, Electron 38.2.2, Chrome 140.0.7339):

```jsonc
// tv_launch via MCP — zero external scripts, zero permission prompts
{
  \"success\": true,
  \"platform\": \"win32\",
  \"launch_method\": \"msix\",
  \"binary\": \"TradingView.Desktop_n534cwy3pjxzj!TradingView.Desktop\",
  \"pid\": 19136,
  \"cdp_port\": 9222,
  \"cdp_url\": \"http://localhost:9222\",
  \"browser\": \"Chrome/140.0.7339.133\",
  \"user_agent\": \"... TradingView/3.0.0 ... Electron/38.2.2 ...\"
}
```

`npm test` — `tv_launch — auto-detect binary (verify path resolution only)` now passes on Windows (was failing before this PR on non-Mac). Rest of e2e suite unaffected by this change.

## Design notes

- **Why a separate `.ps1` instead of inline PowerShell?** The COM activation requires an `Add-Type` C# block. Inline-escaping that through `execSync` + cmd is fragile (quote hell, encoding). A parametrized `.ps1` stays readable, is greppable by users, and keeps the Node code tiny.
- **Why fall back to MSIX, not try it first?** Preserves existing behavior for anyone still on a legacy native install (though those are rare in 2026).
- **Security considerations:** `IApplicationActivationManager` is an OS-provided COM interface — no third-party deps. The only argument passed is `--remote-debugging-port=<port>`, same flag the existing native path uses. No new attack surface.
- **Prior art:** @emremigh published a [working MSIX launcher](https://github.com/emremigh/tradingview-mcp-windows-msix-fix) as an external fix. This PR brings that approach into the core `tv_launch` tool so users don't need a separate script.

## Test plan

- [x] `npm test` — previously failing `tv_launch` test now passes on Windows
- [x] Live `tv_launch` MCP call on Windows 11 MSIX install → CDP ready in a single call
- [ ] Reviewer verification on a native Windows install (legacy path unchanged, should be a no-op)
- [ ] Reviewer verification on macOS / Linux (MSIX branch is win32-only, should be a no-op)

## Fixes

- Closes #14 (Windows App)
- Closes #23 (launch_tv_debug.bat doesn't detect Appx/WindowsApps installs)
- Closes #42 (Windows MSIX: CDP fails with --remote-debugging-port)

### DIFF
--- a/README.md
+++ b/README.md
@@ -100,6 +100,8 @@ TradingView Desktop must be running with Chrome DevTools Protocol enabled on por
 scripts\launch_tv_debug.bat
 ```
 
+> **Windows MSIX note:** TradingView Desktop on Windows ships as an MSIX package (even when downloaded from tradingview.com/desktop/), installed under `C:\Program Files\WindowsApps\`. The Windows sandbox blocks direct `.exe` invocation with CLI flags, so `launch_tv_debug.bat` will not find a native install. Use the MCP tool `tv_launch` instead (or your AI assistant) — it auto-detects the MSIX package and launches via `IApplicationActivationManager`, the official Microsoft COM API for activating packaged apps with arguments.
+
 **Linux:**
 ```bash
 ./scripts/launch_tv_debug_linux.sh
@@ -331,7 +333,8 @@ Launch scripts and `tv_launch` auto-detect TradingView. If auto-detection fails:
 | Platform | Common Locations |
 |----------|-----------------|
 | **Mac** | `/Applications/TradingView.app/Contents/MacOS/TradingView` |
-| **Windows** | `%LOCALAPPDATA%\TradingView\TradingView.exe`, `%PROGRAMFILES%\WindowsApps\TradingView*\TradingView.exe` |
+| **Windows (legacy)** | `%LOCALAPPDATA%\TradingView\TradingView.exe`, `%PROGRAMFILES%\TradingView\TradingView.exe` |
+| **Windows (MSIX)** | Auto-detected via `Get-AppxPackage -Name TradingView.Desktop`; launched via `IApplicationActivationManager` COM API — use `tv_launch`, not `.exe` directly |
 | **Linux** | `/opt/TradingView/tradingview`, `~/.local/share/TradingView/TradingView`, `/snap/tradingview/current/tradingview` |
 
 The key flag: `--remote-debugging-port=9222`
--- a/scripts/launch_msix.ps1
+++ b/scripts/launch_msix.ps1
@@ -1,0 +1,77 @@
+# =====================================================
+#  TradingView Desktop MSIX Debug Launcher (helper)
+#  Used by src/core/health.js -> launch() on Windows
+#  when TradingView is installed as an MSIX package
+#  (Program Files\WindowsApps\), which blocks direct
+#  .exe invocation with --remote-debugging-port=<port>.
+#
+#  Launches via IApplicationActivationManager, the
+#  official Microsoft COM API for activating packaged
+#  apps with arguments.
+#
+#  Emits a single-line JSON result to stdout.
+# =====================================================
+
+[CmdletBinding()]
+param(
+    [Parameter(Mandatory = $true)]
+    [string]$Aumid,
+
+    [Parameter(Mandatory = $false)]
+    [int]$Port = 9222
+)
+
+$ErrorActionPreference = 'Stop'
+
+Add-Type @"
+using System;
+using System.Runtime.InteropServices;
+
+public class TVLauncher {
+    [ComImport]
+    [Guid("2e941141-7f97-4756-ba1d-9decde894a3d")]
+    [InterfaceType(ComInterfaceType.InterfaceIsIUnknown)]
+    public interface IApplicationActivationManager {
+        int ActivateApplication(
+            [MarshalAs(UnmanagedType.LPWStr)] string appUserModelId,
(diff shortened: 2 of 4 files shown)


tradesdontlie:

PR #78, opened 2026-04-18
Fix: draw_clear, draw_list, draw_remove_one, draw_get_properties were…

oager:
## Problem

`src/core/drawing.js` imports its dependencies with underscore-prefixed aliases (`evaluate as _evaluate`, `getChartApi as _getChartApi`) for the DI pattern used by `drawShape`. But four sibling functions — `listDrawings`, `getProperties`, `removeOne`, and `clearAll` — still call the bare names `getChartApi()` and `evaluate()`, which don't exist in scope. Calling any of these throws:

```
ReferenceError: getChartApi is not defined
```

## Reproduction

```
draw_clear  → error: "getChartApi is not defined"
draw_list   → same
draw_remove_one → same
draw_get_properties → same
draw_shape  → works (uses the _resolve(deps) pattern correctly)
```

## Fix

Rename the bare references in the four affected functions to use the imported aliases. `drawShape` is unchanged. Minimal surgical change: 8 insertions / 8 deletions.

## Why existing tests didn't catch this

`drawing.js` currently has no direct test coverage in the `npm test` suite (pine_analyze / cli / e2e don't invoke these paths). I verified the module loads and the fixed functions execute without the ReferenceError.

## Scope

Bug fix only. No public API changes. No behavior change beyond fixing the crash. Single-concern per CONTRIBUTING.md.

### DIFF
--- a/src/core/drawing.js
+++ b/src/core/drawing.js
@@ -45,8 +45,8 @@ export async function drawShape({ shape, point, point2, overrides: overridesRaw,
 }
 
 export async function listDrawings() {
-  const apiPath = await getChartApi();
-  const shapes = await evaluate(`
+  const apiPath = await _getChartApi();
+  const shapes = await _evaluate(`
     (function() {
       var api = ${apiPath};
       var all = api.getAllShapes();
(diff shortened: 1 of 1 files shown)


oager:
Closing this — I missed that #49 and #62 already address the same bug. #62 includes regression tests, so I'd defer to that one. Leaving this here only in case the author wants to draw any of my commit message wording.

PR #90, opened 2026-04-21
fix: TV Desktop 3.1.0 compat for data.trades / data.strategy / data.equity

NickTopel:
## Problem

On TV Desktop 3.1.0 (Electron 38 / Chrome 140), `data_get_trades`, `data_get_strategy_results`, and `data_get_equity` return `trade_count: 0` with `error: "No strategy found on chart"` even when a strategy is clearly attached and Strategy Tester displays thousands of trades.

## Root cause

Two assumptions in `src/core/data.js` no longer hold on TV 3.1.0:

1. **Scrapers filter sources by `metaInfo().is_price_study === false`.** TV 3.1.0 overlay strategies have `is_price_study: true`, so this filter misses them entirely.
2. **Trade / performance / equity data moved from the public `ordersData` / `reportData` / `equityData` accessors to the underscore-prefixed private `_reportData.trades` / `_reportData.performance` / `_reportData.buyHold`.**

The canonical strategy marker is now `metaInfo().id` starting with `StrategyScript` (e.g. `StrategyScript$USER;<uuid>@tv-scripts`) — that's the authoritative "this is a user strategy" flag, independent of overlay behavior.

## Fix

- Added a `__findStrategy(sources)` helper (inlined into each of the three functions' evaluate blocks): prefers the `StrategyScript` metaId marker, falls back to the legacy `is_price_study === false && (ordersData || reportData || _reportData)` filter for older TV builds.
- `getStrategyResults`: prefers `_reportData.performance`, flattens nested `all` / `long` / `short` sub-objects with dotted keys (`all.totalTrades`, `long.netProfit`, etc.). Legacy `reportData` / `performance()` paths still execute when `_reportData` is absent.
- `getTrades`: prefers `_reportData.trades`, emits a normalized shape (`entry_order_id`, `entry_price`, `entry_time_ms`, `entry_type`, `exit_*`, `quantity`, `pnl{_pct}`, `cum_pnl{_pct}`, `runup{_pct}`, `drawdown{_pct}`). Adds `total_trade_count` alongside the (limit-capped) `trade_count`. Legacy `ordersData` / `_orders` / `tradesData` path preserved.
- `getEquity`: prefers `_reportData.buyHold` per-bar equity points, falls back to legacy `equityData`.

## Verification

Tested live on **TV Desktop 3.1.0.7818** with a Bollinger Bands strategy on SPY 5m over 365 days, and with a large VSE strategy on EURUSD 1m over entire history:

| Call | Before | After |
|---|---|---|
| `data trades` | `trade_count: 0`, `error: "No strategy found on chart"` | 454 closed-trade pairs (Bollinger) / 9000 (VSE) with normalized columns |
| `data strategy` | `metric_count: 0` | 106 flattened metrics incl. `sharpeRatio`, `sortinoRatio`, `maxStrategyDrawDown`, `buyHoldReturn`, `all.totalTrades`, `all.profitFactor`, `all.winRate`, plus `long.*` / `short.*` breakdowns |
| `data equity` | `error: "No strategy found on chart"` | 9001 per-bar equity points |

No regression path tested on pre-3.1.0 TV; the legacy code paths are intact and fire when `_reportData` is absent.

## Notes for maintainers

- Zero new dependencies, zero changes to the connection layer, no schema change to tool outputs (only additive fields: `total_trade_count`, dotted-key `all.*` metrics, etc.).
- If you'd prefer the legacy flat-metric shape over dotted nesting for performance keys, happy to adjust.
- The three patched functions share the same `__findStrategy` helper inline — could be extracted to a module-level helper if you prefer, let me know.

Thanks for the excellent tool!

### DIFF
--- a/CLAUDE.md
+++ b/CLAUDE.md
@@ -43,13 +43,27 @@ Use `study_filter` parameter to target a specific indicator by name substring (e
 
 ### "Work on Pine Script"
 1. `pine_set_source` → inject code into editor
-2. `pine_smart_compile` → compile with auto-detection + error check
+2. `pine_smart_compile` → compile with auto-detection + error check. Pass `target: "save" | "add_to_chart" | "save_and_add"` to force the button choice (default = legacy heuristic).
 3. `pine_get_errors` → read compilation errors
 4. `pine_get_console` → read log.info() output
 5. `pine_get_source` → read current code back (WARNING: can be very large for complex scripts)
 6. `pine_save` → save to TradingView cloud
 7. `pine_new` → create blank indicator/strategy/library
 8. `pine_open` → load a saved script by name
+9. `pine_add_to_chart` → click Add-to-Chart and **confirm** study was replaced (polls study list). Use after editing a script that's already on the chart — `pine_smart_compile` may default to Save and leave the chart bound to old code.
+10. `pine_save_and_add_to_chart` → compound (save + dialog dismiss + add-to-chart) in a single round-trip.
+11. `chart_replace_study` → compound: re-add a Pine script with input overrides applied to the new study (eliminates "settings reset to defaults on re-add" friction).
+
+### "Profile a Pine script's execution cost"
+1. `pine_open` (or `pine_set_source` + `pine_smart_compile`) → load the script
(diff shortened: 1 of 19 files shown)


tradesdontlie:

PR #93, opened 2026-04-23
fix: detect MSIX/WindowsApps TradingView install using PowerShell

mkharbai:
## Summary

- The existing `dir /s /b` approach for detecting MSIX/WindowsApps installs silently fails on Windows because `C:\Program Files\WindowsApps` is access-restricted for standard users
- Replaced with a `powershell Get-ChildItem` call that can traverse the restricted directory without elevated privileges
- Verified working on Windows 11 with TradingView installed as an MSIX package via Windows Store

## What changed

`scripts/launch_tv_debug.bat`: replaced `dir /s /b` WindowsApps detection with a PowerShell `Get-ChildItem` call that bypasses the directory access restriction.

## Why

On standard Windows accounts, `C:\Program Files\WindowsApps` cannot be listed with `dir`, causing MSIX detection to silently fail and the script to report TradingView as not found even when installed. PowerShell `Get-ChildItem -ErrorAction SilentlyContinue` resolves this without requiring admin rights.

## Testing

Tested on Windows 11 with TradingView Desktop installed via Windows Store (MSIX). Script now correctly detects and launches TradingView with `--remote-debugging-port=9222`.

Generated with [Claude Code](https://claude.com/claude-code)


tradesdontlie:

PR #94, opened 2026-04-23
chore(cdp): env-var overrides for TV CDP host/port

richroberts-prog:
Adds `TV_CDP_HOST` / `TV_CDP_PORT` env-var support to `src/connection.js` and `src/core/tab.js`, so the MCP can connect to a TV instance on any CDP port.

**Motivation:** Running TV on the default `:9222` clashes with other CDP-enabled processes (Chrome, etc.). Previously the only fix was patching the hardcoded port locally.

**Usage:**
```
TV_CDP_PORT=9223 node src/server.js
# connects to TV launched with --remote-debugging-port=9223
```

Backwards-compatible — defaults to `localhost:9222` when env vars are unset.

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -10,6 +10,9 @@
       "dependencies": {
         "@modelcontextprotocol/sdk": "^1.12.1",
         "chrome-remote-interface": "^0.33.2"
+      },
+      "bin": {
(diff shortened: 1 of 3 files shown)


tradesdontlie:

PR #96, opened 2026-04-23
fix(data): DOM-scrape fallback for strategy results + trades

richroberts-prog:
`getStrategyResults()` and `getTrades()` used `s.reportData` / `s.ordersData` off `dataSources()`. In recent TV builds this internal API returns empty objects even when a strategy is live on chart and the Strategy Report panel shows populated data.

**Fix:** DOM-scrape fallback parses the visible Strategy Report panel. Internal API stays as the fast path. Response now includes `source: 'internal_api' | 'dom_fallback'` so callers can observe which path served the data.

**Tested on:** TV Desktop 3.1.0 (Electron 38.2.2), macOS. Exercised against a default SMA-crossover strategy on FX:EURUSD 1h. Internal API returned empty; DOM fallback returned numbers matching the visible Strategy Tester panel byte-for-byte (Total P&L, PF, trade count, max DD).

**Regression risk:** low — DOM path only runs when internal API returns falsy; no change to the existing code path when the API works.

### DIFF
--- a/src/core/data.js
+++ b/src/core/data.js
@@ -135,6 +135,7 @@ export async function getIndicator({ entity_id }) {
 export async function getStrategyResults() {
   const results = await evaluate(`
     (function() {
+      // Primary path: internal TV API. Has drifted across TV versions; kept as fast-path.
       try {
         var chart = ${CHART_API}._chartWidget;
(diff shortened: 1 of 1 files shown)


tradesdontlie:

PR #100, opened 2026-04-24
fix(launch): detect TradingView Microsoft Store install on Windows

kirillbobrihin-cyber:
The tv_launch tool only searched standard install paths on Windows: %LOCALAPPDATA%\TradingView\, Program Files, and Program Files (x86). It did not handle the Microsoft Store version, which lives in a versioned WindowsApps folder inaccessible via existsSync.

Add two fallback detection methods for Windows Store installs:
1. Query the path of a running TradingView process via PowerShell (Get-Process | Select Path) — works when app is already open.
2. Query the install location via Get-AppxPackage — works even when the app is not running. Skips existsSync since WindowsApps requires admin rights to read.

### DIFF
--- a/src/core/health.js
+++ b/src/core/health.js
@@ -197,6 +197,30 @@ export async function launch({ port, kill_existing } = {}) {
     } catch { /* ignore */ }
   }
 
+  // Windows Microsoft Store version: lives in a versioned WindowsApps folder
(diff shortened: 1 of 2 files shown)


ceprateek:
thanks, very much needed

tradesdontlie:

PR #104, opened 2026-04-25
fix(quote_get): honor symbol parameter via chart switch + restore

ceprateek:
## Summary

`quote_get`'s `symbol` parameter was accepted but ignored for data — it was only stamped onto the returned object as a label. The IIFE always read OHLC from the active chart's `bars()`, so e.g. `quote_get('NVDA')` while the chart was on TSLA returned TSLA's OHLC with `symbol: 'NVDA'` glued on top. Calling parallel `quote_get` across many symbols therefore returned identical data for all of them.

This PR makes `quote_get` actually honor the `symbol` parameter.

## Implementation

- When a symbol is provided and differs from the current chart symbol (compared by bare ticker, after exchange prefix), `getQuote` now calls `chart.setSymbol`, awaits `waitForChartReady`, reads the quote, and restores the original symbol in a `finally` block so errors don't leave the chart on the wrong ticker.
- A module-level Promise-chain lock (`_quoteLock`) serializes parallel callers so concurrent `quote_get(A)` / `quote_get(B)` don't race over the shared chart state.
- Same-symbol calls (no switch needed) skip the switch entirely — zero added latency.
- Tool description in `tools/data.js` is updated so callers know the cross-symbol path costs ~1-2s and serializes.

## Tradeoffs

- ~1-2s overhead per cross-symbol quote (chart load + restore).
- Cross-symbol calls serialize through the lock — no parallel speedup, but data is finally correct.
- A truly non-mutating implementation exists at `_chartApiInstance._sessions.qs_snapshoter_basic-symbol-quotes_*` (TV's internal WebSocket snapshot service), but reverse-engineering its protocol is a much larger change. Left as future work.

## Test plan

- [x] Smoke test: 4 parallel `quote_get` calls (AAPL, NVDA, META, AMD) from a chart on MSFT — each returned distinct correct OHLC + description; chart restored to MSFT after.
- [x] Same-symbol fast-path: `quote_get` with the chart's current symbol returns immediately without switching.
- [x] Existing e2e tests: the embedded IIFE is unchanged in shape (only the `sym` resolution moved out, since the chart is now guaranteed to be on the right symbol when the IIFE runs); size-limit test stays under 500 bytes.
- [ ] CI on upstream.

### DIFF
--- a/src/core/data.js
+++ b/src/core/data.js
@@ -2,11 +2,18 @@
  * Core data access logic.
  */
 import { evaluate, evaluateAsync, KNOWN_PATHS, safeString } from '../connection.js';
+import { waitForChartReady } from '../wait.js';
 
 const MAX_OHLCV_BARS = 500;
 const MAX_TRADES = 20;
 const CHART_API = KNOWN_PATHS.chartApi;
 const BARS_PATH = KNOWN_PATHS.mainSeriesBars;
+
+// Serializes getQuote() calls that mutate chart symbol so concurrent callers
+// can't race over the shared chart state. JS is single-threaded but our
+// awaits interleave; without this every parallel quote_get(symbol) would
+// read whichever symbol the chart happened to be on at evaluate() time.
+let _quoteLock = Promise.resolve();
 
 function buildGraphicsJS(collectionName, mapKey, filter) {
   return `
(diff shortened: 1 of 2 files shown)


tradesdontlie:
Verified live before merging on TradingView Desktop 3.1.0: quote_get(symbol=AAPL) from an NQ1! chart returned genuine Apple data (BATS:AAPL, correct description/type), the chart restored to CME_MINI:NQ1! automatically, and the no-symbol fast path is unchanged. The mislabeled-data failure mode (NQ prices stamped with the requested symbol name) is gone. Appreciated the honest tool-description note about the switch cost and the concurrency lock. Thanks @ceprateek!

PR #109, opened 2026-04-26
Add .claude/ to .gitignore and sync package-lock.json bin entry

Boult14:
## Summary

- Add `.claude/` to `.gitignore` — this directory is created by Claude Code and contains local session/settings files that should never be committed (analogous to `.idea/` or `.vscode/`)
- Sync `package-lock.json` with the `bin` field already declared in `package.json` — running `npm install` revealed this entry was missing from the lockfile

## Test plan

- [ ] Confirm `.claude/` no longer appears as an untracked file after `npm install` in a fresh clone
- [ ] Confirm `npm install` produces no lockfile diff on a clean checkout

🤖 Generated with [Claude Code](https://claude.ai/claude-code)

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -4,3 +4,4 @@ scripts/current.pine
 temp_*
 *.log
 discovery-log.json
+.claude/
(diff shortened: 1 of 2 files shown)


tradesdontlie:

PR #110, opened 2026-04-26
Support MSIX/Microsoft Store TradingView installs on Windows

TmanOKC2026:
The Windows path probes only knew about classic Win32 installs under LOCALAPPDATA / Program Files. Users who installed TradingView Desktop from the Microsoft Store get an MSIX-packaged copy under %PROGRAMFILES%\WindowsApps, which (a) `where TradingView.exe` cannot discover and (b) cannot be launched via the bare .exe path with custom CLI args because of MSIX activation restrictions.

Changes:
- After the existing path probes fail on win32, fall back to Get-AppxPackage to discover an MSIX TradingView install and resolve its AUMID from AppxManifest.xml.
- Launch MSIX-packaged TradingView through a shortcut whose TargetPath is shell:AppsFolder\<AUMID> and whose Arguments carries --remote-debugging-port. This is the only reliable mechanism to pass CLI args through the MSIX activation manager.
- Probe CDP readiness via 127.0.0.1 instead of localhost. On Windows, localhost resolves to ::1 first but Electron only binds the debug port on IPv4, causing the readiness loop to time out even when CDP is up.
- Add scripts/launch_tv_debug.ps1 as a standalone launcher that uses the same MSIX path, useful for users who want to start TradingView with CDP enabled outside of the MCP server.

### DIFF
--- a/scripts/launch_tv_debug.ps1
+++ b/scripts/launch_tv_debug.ps1
@@ -1,0 +1,62 @@
+# Launch TradingView Desktop on Windows with Chrome DevTools Protocol enabled.
+# Handles both classic Win32 installs and MSIX/UWP (Microsoft Store) installs.
+# Usage: powershell -ExecutionPolicy Bypass -File scripts\launch_tv_debug.ps1 [-Port 9222]
+
+param([int]$Port = 9222)
+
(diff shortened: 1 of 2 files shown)


tradesdontlie:

PR #125, opened 2026-05-04
Fix formatting in README.md header

soryd0607:


tradesdontlie:
Closing — the diff only adds a trailing space after the header text, which linters flag rather than fix anything. Thanks for the interest in the project though!

PR #133, opened 2026-05-08
Fix CDP bridge: target picker, liveness probe, symbol resolution

rodaguilera:
## Summary

Cluster of bugs in the CDP bridge that produced inconsistent attach behavior, false readiness signals, and wrong-tab screenshots. All issues were diagnosed and live-validated against TradingView Desktop on macOS.

The bugs compounded silently — `chart_set_symbol` would error out on the wrong tab, `capture_screenshot` would return non-chart content, and `waitForChartReady` could either false-positive (wrong tab, fragile DOM signal) or false-negative timeout (TV resolving exchange prefix). User-facing symptom: the TV chart appeared "blank" or operations seemed to silently no-op when in fact the picker had attached to a news-flow / watchlist / symbols tab.

### 1. `findChartTarget` silent fallback to non-chart tabs (`src/connection.js`)

Previously fell back to any TradingView page if no `/chart/` URL existed:
```js
return targets.find(t => /tradingview\.com\/chart/i.test(t.url))
  || targets.find(t => /tradingview/i.test(t.url))   // ← attaches to anything
  || null;
```
Result: every chart-API tool throws `TypeError: Cannot read properties of undefined (reading 'value')` because `_activeChartWidgetWV` is `undefined` on news-flow / watchlist / symbols pages. Worse: `capture_screenshot` returns the wrong page's content.

**Fix:** fail loud with an actionable error listing the non-chart tabs detected.

### 2. `getClient` cached client never invalidated when stale (`src/connection.js`)

`Runtime.evaluate('1')` passes on any TradingView page with a live JS engine. Once the picker attached to a non-chart tab, opening a chart tab afterward did not refresh the cache.

**Fix:** strict liveness probe checks `window.TradingViewApi._activeChartWidgetWV !== undefined`. Cached clients on tabs without chart APIs are invalidated, forcing reconnect through the (now strict) picker.

### 3. `symbolSearch` `.slice()` throws on `{}` response (`src/core/chart.js`)

`(data.symbols || data || []).slice(0, 15)` evaluates `{}` as truthy and lets `.slice()` throw if the API ever returns an object without a `symbols` field (schema drift / no matches). 

**Fix:** explicit `Array.isArray` checks.

### 4. `waitForChartReady` three readiness bugs (`src/wait.js`)

a. `[class*="bar"]` matched toolbars / sidebars / scrollbars producing a near-constant bar counter; `stableCount` reached threshold in ~400ms regardless of actual chart load. **Fix:** use TV's internal `mainSeries().bars()` API (already known via `KNOWN_PATHS.mainSeriesBars`).

b. Symbol match short-circuited when the legend was empty (legend not yet rendered after symbol change), letting readiness pass before the new symbol was confirmed. **Fix:** wait for non-empty `currentSymbol`.

c. `.includes(expectedSymbol)` is one-way — when caller passes `"NASDAQ:IREN"` but TV resolves to `"BATS:IREN"` (preferred feed), the match never succeeds and timeout fires. **Fix:** compare against bare ticker (segment after last `:`).

## Test plan

Validated end-to-end on TradingView Desktop on macOS:

- [x] `tv_health_check` reports the correct chart tab + `api_available: true` even when other TradingView tabs (news-flow, watchlist) are also open
- [x] `chart_set_symbol("AAPL")` returns `chart_ready: true`; subsequent `chart_get_state` and `quote_get` reflect the new symbol
- [x] `chart_set_symbol("NASDAQ:IREN")` returns `chart_ready: true` (previously timed out due to TV resolving to `BATS:IREN`)
- [x] `capture_screenshot` returns the actual chart, not whatever non-chart tab the old picker happened to attach to
- [x] `symbol_search("ZZZQQQNOTAREALSYMBOL")` returns `count: 0` gracefully
- [x] `node --check` passes on all 3 modified files

Existing automated e2e tests (`tests/e2e.test.js`) not run against this branch — they require a maintainer's test harness configuration. Happy to adapt if there's a CI pattern you'd like me to follow.

### DIFF
--- a/src/connection.js
+++ b/src/connection.js
@@ -53,13 +53,19 @@ export function requireFinite(value, name) {
 export async function getClient() {
   if (client) {
     try {
-      // Quick liveness check
-      await client.Runtime.evaluate({ expression: '1', returnByValue: true });
-      return client;
-    } catch {
-      client = null;
-      targetInfo = null;
-    }
+      // Strict liveness: cached tab must be alive AND expose chart APIs.
+      // Plain `Runtime.evaluate('1')` passes on any TradingView page
+      // (news-flow, watchlist, symbols) and used to lock the picker to
+      // whichever tab was attached first — silently breaking every
+      // chart-API tool when a chart tab opened later.
+      const probe = await client.Runtime.evaluate({
+        expression: 'typeof window.TradingViewApi !== "undefined" && window.TradingViewApi._activeChartWidgetWV !== undefined',
+        returnByValue: true,
+      });
+      if (probe?.result?.value === true) return client;
+    } catch {}
+    client = null;
+    targetInfo = null;
   }
   return connect();
 }
@@ -69,10 +75,10 @@ export async function connect(targetId = null) {
   for (let attempt = 0; attempt < MAX_RETRIES; attempt++) {
     try {
       const target = targetId ? await findTargetById(targetId) : await findChartTarget();
-      if (!target) {
-        throw new Error(targetId
-          ? `CDP target ${targetId} not found — is the tab still open?`
-          : 'No TradingView chart target found. Is TradingView open with a chart?');
+      // findChartTarget now throws explicit errors with actionable hints (no
+      // chart tab vs no TV at all), so only findTargetById can still return null.
+      if (targetId && !target) {
+        throw new Error(`CDP target ${targetId} not found — is the tab still open?`);
       }
       targetInfo = target;
       client = await CDP({ host: CDP_HOST, port: CDP_PORT, target: target.id });
@@ -85,6 +91,11 @@ export async function connect(targetId = null) {
       return client;
     } catch (err) {
       lastError = err;
+      // Don't retry actionable user-facing errors; they won't fix themselves
+      // by waiting (need user to open a chart tab).
+      if (err.message && /No TradingView chart tab found/.test(err.message)) {
+        throw err;
+      }
       const delay = Math.min(BASE_DELAY * Math.pow(2, attempt), 30000);
       await new Promise(r => setTimeout(r, delay));
     }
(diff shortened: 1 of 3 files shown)


rodaguilera:
Rebased onto current `main` — this is no longer conflicting.

The three original fixes are all still needed: `getClient()`, `findChartTarget()`,
`symbolSearch()` and `waitForChartReady()` on `main` today are byte-identical to what
this PR replaces. The July CDP work (`tab_switch` re-attach, configurable host/port)
touched the surrounding code, which is where the conflict came from, not the fixes
themselves.

**Conflict resolution.** `connect()` gained `targetId`, so the `if (!target)` guard I
removed now serves two paths. Kept it for `findTargetById` (which can still return
null) and dropped it only for `findChartTarget` (which now throws).

**One new commit**, because `reconnectTo()` did not exist when this PR was opened:
`getClient()` recovers a dead client with `connect()` and no target, so it re-picks
whatever `findChartTarget()` sees first rather than the tab `tab_switch` selected —
with two or more chart tabs open that silently moves the session to a different chart.
It now tracks the pinned target and returns to it. The commit also exempts a pin from
the strict probe; both of today's `reconnectTo()` callers pin chart tabs, so that half
is defensive only, and it is labelled as such.

**Measured against a live TradingView Desktop**, since the original PR only argued
these from reading the code:

- `document.querySelectorAll('[class*="bar"]')` returns **100** elements on a loaded
  chart — 44 `widgetbar`, 25 `control`, 6 scrollbar, 4 toolbar, and so on. **Zero**
  price bars. The chart had **300** bars in `mainSeries().bars()`. Since UI chrome is
  constant, `stableCount` reaches 2 immediately and `waitForChartReady` reports ready
  regardless of whether data has loaded.
- On `/news-flow/` and `/fundamental-graphs/`, `typeof window.TradingViewApi` is
  **`"object"`** while `_activeChartWidgetWV` is **undefined**. So a laxer probe
  checking only that `TradingViewApi` exists still attaches to those pages — the
  `_activeChartWidgetWV` half is what actually discriminates.
- A chart opened as `?symbol=NYSE%3AGLW` reports `BATS:GLW` from `chart.symbol()`.
  The previous `currentSymbol.includes(expectedSymbol)` never matches that, so
  `waitForChartReady` spins to timeout; comparing the bare ticker resolves it.
- With no chart tab open, `tv_health_check` now returns the actionable error naming
  the non-chart tabs it found, instead of attaching to one of them.

`npm run lint` is clean (4 pre-existing warnings, none in these files) and
`npm run test:unit` passes 152/152. `tests/e2e.test.js` needs a chart tab and drives
the live app, so I have not run it to completion.

CI has not reported — I think the workflow needs your approval to run on a fork PR.

Happy to split the fourth commit into its own PR if you would rather keep this one
scoped to the original three.

Issue #128, opened 2026-05-05
Windows MSIX TradingView build doesn't expose CDP — tv_launch / launch_tv_debug.bat only target legacy non-MSIX paths


HussamZeibak:
## Summary

On Windows 11, the only TradingView Desktop distribution available now appears to be an **MSIX package** (installs to `C:\Program Files\WindowsApps\TradingView.Desktop_<hash>\`). This build accepts `--remote-debugging-port=9222` on its command line but **never binds the port** — Chromium devtools is filtered out before Chromium parses it. None of the launcher mechanisms in the repo work against it.

## Environment

- Windows 11 Home 26200
- TradingView MSIX: `TradingView.Desktop_3.1.0.7818_x64__n534cwy3pjxzj` (`Get-AppxPackage` `SignatureKind: Developer`, `InstallLocation: C:\Program Files\WindowsApps\...`)
- Node v24.13.1, npm 11.8.0
- Cloned `main` (latest commit at the time of testing)

## What I tried

| Method | Args reached `Win32_Process.CommandLine`? | Port 9222 bound? |
|---|---|---|
| Direct `.exe` from `WindowsApps` (`Start-Process` / `& exe`) | n/a — process exits immediately (MSIX identity required) | No |
| `IApplicationActivationManager::ActivateApplication` (proper COM activation, args param) | Yes — main proc cmdline shows `--remote-debugging-port=9222`, all renderers inherit it | No |
| `scripts/launch_tv_debug.vbs` (`ELECTRON_EXTRA_LAUNCH_ARGS` + `explorer.exe shell:AppsFolder\<AUMID>`) | No — main proc cmdline empty | No |
| `scripts/launch_tv_debug.bat` (legacy `start "" "%TV_EXE%"`) | n/a — same direct-exec failure | No |

The COM-activation result is the conclusive one: the flag literally appeared in `Win32_Process.CommandLine` of the main process (PID confirmed via `Get-CimInstance`), every renderer subprocess inherited it, **and `Get-NetTCPConnection -LocalPort 9222` still returned nothing**. The Electron main script appears to strip Chromium debug switches before Chromium parses the command line.

## Code paths affected

- `src/core/health.js` `launch()` (lines ~172-176): the Windows `pathMap` only lists `%LOCALAPPDATA%\TradingView\`, `%PROGRAMFILES%\TradingView\`, `%PROGRAMFILES(X86)%\TradingView\` — the MSIX install location is never searched.
[...]

PR #135, opened 2026-05-10
fix: support Windows Store (MSIX) TradingView launch and patch npm vulns

vladafedyanin-lang:
### scripts/launch_tv_debug.bat - MSIX launch fix
Windows Store installs of TradingView live under `C:\Program Files\WindowsApps`, which has ACL restrictions that silently block both direct exe execution and the existing `dir /s /b` discovery. The script was exiting without any visible error.

**Fix:** when TradingView.exe is not found in standard paths, fall back to launching via the COM `IApplicationActivationManager` interface (CLSID `45BA127D-10A8-46EA-8AB7-56EA9078943C`) through an inline PowerShell one-liner. This is the same mechanism the Windows Start Menu uses and correctly forwards `--remote-debugging-port` into the Electron process.

Verified working: CDP responds at `http://localhost:9222/json/version` with Electron 38 / Chrome 140.

### `package-lock.json` - security fixes
Ran `npm audit fix` to resolve 5 vulnerabilities (1 high, 4 moderate):
- `fast-uri` ?3.1.1 ? 3.1.2 (path traversal, host confusion)
- `hono` ?4.12.17 ? 4.12.18 (cookie handling, JSX injection, cache leakage, JWT validation)
- `@hono/node-server` <1.19.13 ? 1.19.14 (middleware bypass)
- `ip-address` ?10.1.0 ? 10.2.0 (XSS)
- `express-rate-limit` 8.3.1 ? 8.5.1 (depends on ip-address)

## Testing
- Launched TradingView Desktop (MSIX, Windows Store) via the updated script - CDP confirmed live at `ws://localhost:9222`
- `npm audit` reports 0 vulnerabilities after fix

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -10,12 +10,15 @@
       "dependencies": {
         "@modelcontextprotocol/sdk": "^1.12.1",
         "chrome-remote-interface": "^0.33.2"
+      },
+      "bin": {
+        "tv": "src/cli/index.js"
       }
     },
     "node_modules/@hono/node-server": {
-      "version": "1.19.11",
-      "resolved": "https://registry.npmjs.org/@hono/node-server/-/node-server-1.19.11.tgz",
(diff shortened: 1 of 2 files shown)


tradesdontlie:

PR #148, opened 2026-05-13
fix: wait for chart render before screenshot

### LINKED ISSUE #144: capture_screenshot returns stale frame after chart_set_symbol / chart_set_timeframe
## Description

After calling `chart_set_symbol` or `chart_set_timeframe`, an immediate `capture_screenshot` (region=`chart` or `full`, method=`cdp`) returns an image rendered from the **previous chart state**. The data layer (`chart_get_state`, `quote_get`, `data_get_study_values`, `data_get_pine_tables`) updates correctly and reflects the new symbol/timeframe, but the visible canvas in the screenshot is the previous symbol's rendered frame.

This is an observation rather than a hard bug — the timing window varies (sometimes 0 seconds, sometimes 5+ seconds) and seems related to TradingView's own canvas re-render latency rather than MCP itself. But it consistently causes user-visible issues when an LLM tries to combine screenshot-based pattern recognition with data-layer values within the same analytical session.

## Steps to reproduce

1. Start with chart on `BATS:MSTR`, resolution `60`.
2. Call `chart_set_symbol(symbol="BITSTAMP:BTCUSD")`.
3. Call `chart_set_timeframe(timeframe="D")`.
4. Immediately call `capture_screenshot(region="chart")`.
5. Inspect the returned PNG.

## Expected behavior

Screenshot shows BTCUSD daily candles, BTC OHLC in the header bar.

## Actual behavior

Screenshot shows MSTR 1H candles, MSTR OHLC in the header (e.g. `Strategy Inc Class A · 1H · NASDAQ` — locale-dependent string), even though `chart_get_state` confirms `symbol: "BITSTAMP:BTCUSD"` and `resolution: "1D"`.

A second `capture_screenshot` 1-2 seconds later usually returns the correct BTC frame, but not always. Observed at least 8 times on 2026-05-11 across BTC/MSTR/MO switches.

## Environment

- macOS: Darwin 25.4.0
- TradingView Desktop: stable channel
- TV plan: Essential
- MCP fork: `JDuckling/tradingview-mcp` (likely upstream too)
- Date observed: 2026-05-11

## Workaround

Use `chart_get_state` to verify the symbol/timeframe before relying on the screenshot. If a visual match is required, retry `capture_screenshot` after a short delay (1-3 seconds), or skip the screenshot entirely and rely on data-layer tools (`data_get_study_values`, `data_get_pine_tables`, `quote_get`) which return correct values immediately.

For the `cdp` capture method specifically — observed lag is more pronounced than for `api` method (TradingView's own screenshot mechanism). However `api` method doesn't return file content directly, only triggers TV's UI to save/show.

## Severity & impact

**Low.** Workaround is simple (verify state via data layer, retry capture, or skip screenshot). However, it's a footgun for any LLM-based workflow that tries to do vision-based pattern recognition right after a chart-switch — the result will be a confident analysis of the WRONG instrument.

## Suggested fix direction (without prescription)

Probably MCP could add an optional `wait_for_render` flag (or a short blocking sleep with TV's `tradingview.charts[0].onSymbolChanged` event) to `capture_screenshot` that waits for the canvas to finish re-rendering before returning. Even a fixed 500-1000ms post-switch delay would help most cases.

Alternative: expose a `chart_ready_wait()` helper tool that returns when TV's chart is fully rendered for the current state.

## Related

This is a known observation in our internal `2026-05-11-tv-workspace-layout-watchlist.md` plan reflection; affected multiple analytical sessions throughout the day where Claude reported chart-content discrepancies between visual screenshot and data-layer values.

tlcreativeart-hub:
## Summary

Fixes #144.

This adds an optional `wait_for_render` parameter to `capture_screenshot`.

When enabled, the screenshot tool waits for the visible chart canvas to stabilize before capturing. This helps avoid stale screenshots immediately after `chart_set_symbol` or `chart_set_timeframe`, where the data layer may already reflect the new symbol/timeframe but the visible chart canvas can still show the previous rendered frame.

## Changes

- Added `wait_for_render` option to `capture_screenshot`
- Added `waitForChartRender()` helper in `src/wait.js`
- Included `waited_for_render` in screenshot results for easier debugging
- Kept the existing default behavior unchanged

## Testing

Manual testing recommended:

1. Start on `BATS:MSTR`, resolution `60`
2. Call `chart_set_symbol` with `BITSTAMP:BTCUSD`
3. Call `chart_set_timeframe` with `D`
4. Call `capture_screenshot` with `region: "chart"` and `wait_for_render: true`
5. Confirm the screenshot shows BTCUSD daily chart rather than the previous MSTR 1H frame

AI assistance was used to inspect the issue and draft the fix, and the final changes were reviewed before submission.

### DIFF
--- a/src/core/capture.js
+++ b/src/core/capture.js
@@ -1,7 +1,8 @@
-/**
+﻿/**
  * Core screenshot/capture logic.
  */
 import { getClient, evaluate, getChartCollection } from '../connection.js';
+import { waitForChartRender } from '../wait.js';
 import { writeFileSync, mkdirSync } from 'fs';
 import { join, dirname } from 'path';
 import { fileURLToPath } from 'url';
@@ -9,11 +10,15 @@ import { fileURLToPath } from 'url';
 const __dirname = dirname(fileURLToPath(import.meta.url));
 const SCREENSHOT_DIR = join(dirname(dirname(__dirname)), 'screenshots');
 
-export async function captureScreenshot({ region, filename, method } = {}) {
+export async function captureScreenshot({ region, filename, method, waitForRender = false } = {}) {
   mkdirSync(SCREENSHOT_DIR, { recursive: true });
 
+  if (waitForRender) {
+    await waitForChartRender();
+  }
+
   const ts = new Date().toISOString().replace(/[:.]/g, '-');
-  const fname = (filename || `tv_${region}_${ts}`).replace(/[\/\\]/g, '_');
+  const fname = (filename || `tv_${region || 'full'}_${ts}`).replace(/[\\/]/g, '_');
   const filePath = join(SCREENSHOT_DIR, `${fname}.png`);
 
   if (method === 'api') {
@@ -21,7 +26,9 @@ export async function captureScreenshot({ region, filename, method } = {}) {
       const colPath = await getChartCollection();
       await evaluate(`${colPath}.takeScreenshot()`);
       return {
-        success: true, method: 'api',
+        success: true,
+        method: 'api',
+        waited_for_render: !!waitForRender,
         note: 'takeScreenshot() triggered — TradingView will save/show the screenshot via its own UI',
       };
     } catch {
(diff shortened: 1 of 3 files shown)


tradesdontlie:
Thanks @tlcreativeart-hub — this was the right fix and I verified it live (symbol switch → immediate capture returns a fresh frame). It stopped merging cleanly after the path-traversal fix (#177) touched the same lines, so I re-applied it against current main with you credited in the commit. Landed in #326.

PR #160, opened 2026-05-16
Fix tab_switch to re-attach CDP client to the new target

bbem1985:
switchTab() previously only called /json/activate/{id}, which brings a tab to the front in the UI but does not change which target the CDP client is bound to. As a result, after switching tabs, all reads (chart_get_state, data_get_study_values, quote_get, etc.) still came from whichever tab was first connected to — making per-tab dashboards or cross-tab batch sweeps impossible.

This change adds an optional targetId to connect() and a new reconnectTo(targetId) helper. switchTab() now activates the tab and then re-attaches the cached CDP client to that target, so subsequent reads come from the switched-to tab.

### DIFF
--- a/src/connection.js
+++ b/src/connection.js
@@ -61,13 +61,19 @@ export async function getClient() {
   return connect();
 }
 
-export async function connect() {
+export async function connect(targetId = null) {
   let lastError;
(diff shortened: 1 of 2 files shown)


tradesdontlie:

PR #169, opened 2026-05-19
Fix alert_create stale [class*=alert] selector after TV dialog refactor

### LINKED ISSUE #168: Fix alert_create price_set:false — stale [class*="alert"] selector after TV dialog refactor
## Symptom

`alert_create` returns `success: false, price_set: false, source: "dom_fallback"` against the current TradingView Desktop UI. The alert dialog visibly opens with the symbol's current price pre-filled, but the user-supplied `price` argument is never written into the input. No alert is created.

Reproducible across all condition values (`crossing`, `less_than`, `crossing_down`).

## Repro

1. Launch TV with `scripts/launch_tv_debug_mac.sh`, switch chart to any symbol (e.g. NVDA daily)
2. Call `alert_create { condition: "crossing", price: 215, message: "test" }`
3. Observe: dialog opens (filled with current price ~222.32), dialog stays open with the wrong value, response is `{ success: false, price_set: false }`
4. `alert_list` confirms no alert was created

## Root cause

`src/core/alerts.js:26`:

```js
var inputs = document.querySelectorAll('[class*="alert"] input[type="text"], [class*="alert"] input[type="number"]');
```

This selector requires the price input to be inside an element whose class contains the substring `alert`. In a recent TradingView Desktop update, the alert-creation dialog wrapper was changed and **no longer has any `data-name`, and none of its class fragments contain `alert`**. The wrapper now looks like:

```html
<div id="overlap-manager-root">
  <div class="dialog-qyCw0PaN dialog-YKU5b5xj popup-LEkd5gPO dialog-aRAWUDhF rounded-aRAWUDhF shadowed-aRAWUDhF">
    <form class="form-h6NNXQD2">
      ...
      <div class="condition-Q92ASNAn">
        <fieldset class="container-sFcMHof4">
          ...
          <div class="bandsRow-Q92ASNAn">
            <div class="row-Ur64z4gH">
              <div class="bandsRange-rNHjswTO">
                <input value="222.32" class="input-RUSovanF ...">  <!-- the price input -->
```

Verified by injecting JS via the CDP bridge: `document.querySelectorAll('[class*="alert"] input')` returns 0 nodes when the dialog is open. The dialog DOES open (`set-alert-button` click works fine), and the React-controlled input DOES accept the existing native-setter approach in the function — only the selector that finds the input is wrong.

## Working fix (proven)

The most stable anchor inside the new dialog is the `Create` submit button (text-matched). Locate the dialog by walking up from that button, then find the price input as the first visible numeric-valued input within the dialog. This is more resilient than class-fragment matching because TV's hashed CSS class names rotate per release.

Verified working end-to-end:

```js
const createBtn = Array.from(document.querySelectorAll('button'))
  .find(b => /^create$/i.test(b.textContent.trim()) && b.offsetParent !== null);
const dialog = createBtn?.closest('div[class*="dialog"]');
const inputs = Array.from(dialog.querySelectorAll('input')).filter(i => i.offsetParent !== null);
const priceInput = inputs.find(i => !isNaN(parseFloat(i.value)));
// then the existing native React setter trick in alerts.js works as-is
```

I'll open a PR with this fix + an e2e test covering the new selector strategy.

## Environment

- TradingView Desktop on macOS (Darwin 25.4.0, Apple Silicon)
- tradingview-mcp commit `4795784` (current `main`)
- TV Basic plan account
- Date observed: 2026-05-19

amhamid:
## Summary

Fixes #168.

`alert_create` silently returned `success: false, price_set: false` against the current TradingView Desktop UI: the dialog opened with the symbol's current price pre-filled, the user-supplied `price` was never written, and no alert was created.

Root cause: `src/core/alerts.js:26` required the price input to live inside an ancestor whose class fragment contains `alert`, but a recent TV update changed the dialog wrapper so it has no `data-name` and no class fragment containing `alert`. `[class*="alert"] input` matches 0 nodes when the dialog is open.

## Fix

Anchor on the `Create` submit button (text-matched, stable across releases), walk up to its dialog ancestor, then find the price input as the first visible numeric-valued input inside the dialog. The native React `value` setter and submit-button click are unchanged.

```diff
- var inputs = document.querySelectorAll('[class*="alert"] input[type="text"], [class*="alert"] input[type="number"]');
- // ...label-text matching that no longer matches anything...
+ var createBtn = Array.from(document.querySelectorAll('button'))
+   .find(function(b) { return /^create$/i.test(b.textContent.trim()) && b.offsetParent !== null; });
+ var dialog = createBtn && createBtn.closest('div[class*="dialog"]');
+ if (!dialog) return false;
+ var inputs = Array.from(dialog.querySelectorAll('input')).filter(function(i) { return i.offsetParent !== null; });
+ var priceInput = inputs.find(function(i) { return !isNaN(parseFloat(i.value)); });
```

## Test plan

- [x] Added `tests/e2e.test.js` regression test in the `Alerts` describe block that opens the dialog and verifies the new selector strategy finds a numeric price input. Cleans up via Escape.
- [x] `npm test`: alert suite 4/4 passing (3 pre-existing + 1 new). The 2 unrelated failures in `ui_open_panel` and `replay_stop` exist on `main` (TV API drift / replay state) and are out of scope here.
- [x] Live-tested against TradingView Desktop on macOS — the same JS body now in `alerts.js` was used to successfully create alert `4734304563` (NVDA crossing 215) during diagnosis.

## Scope

One focused change. No behavior change beyond the selector swap. Sibling functions `list` and `deleteAlerts` use different code paths and are not touched.

### DIFF
--- a/src/core/alerts.js
+++ b/src/core/alerts.js
@@ -23,24 +23,22 @@ export async function create({ condition, price, message }) {
 
   const priceSet = await evaluate(`
     (function() {
-      var inputs = document.querySelectorAll('[class*="alert"] input[type="text"], [class*="alert"] input[type="number"]');
-      for (var i = 0; i < inputs.length; i++) {
-        var label = inputs[i].closest('[class*="row"]')?.querySelector('[class*="label"]');
-        if (label && /value|price/i.test(label.textContent)) {
-          var nativeSet = Object.getOwnPropertyDescriptor(HTMLInputElement.prototype, 'value').set;
-          nativeSet.call(inputs[i], ${safeString(String(price))});
-          inputs[i].dispatchEvent(new Event('input', { bubbles: true }));
-          inputs[i].dispatchEvent(new Event('change', { bubbles: true }));
-          return true;
-        }
-      }
-      if (inputs.length > 0) {
-        var nativeSet = Object.getOwnPropertyDescriptor(HTMLInputElement.prototype, 'value').set;
-        nativeSet.call(inputs[0], ${safeString(String(price))});
-        inputs[0].dispatchEvent(new Event('input', { bubbles: true }));
-        return true;
-      }
-      return false;
+      // The alert dialog wrapper has no data-name and no class fragment containing "alert"
+      // in current TradingView builds, so [class*="alert"] returns 0 inputs (see issue #168).
+      // Anchor on the "Create" submit button instead — its text is stable across releases —
+      // then find the price input as the first visible numeric-valued input in that dialog.
+      var createBtn = Array.from(document.querySelectorAll('button'))
+        .find(function(b) { return /^create$/i.test(b.textContent.trim()) && b.offsetParent !== null; });
+      var dialog = createBtn && createBtn.closest('div[class*="dialog"]');
+      if (!dialog) return false;
+      var inputs = Array.from(dialog.querySelectorAll('input')).filter(function(i) { return i.offsetParent !== null; });
+      var priceInput = inputs.find(function(i) { return !isNaN(parseFloat(i.value)); });
+      if (!priceInput) return false;
+      var nativeSet = Object.getOwnPropertyDescriptor(HTMLInputElement.prototype, 'value').set;
+      nativeSet.call(priceInput, ${safeString(String(price))});
+      priceInput.dispatchEvent(new Event('input', { bubbles: true }));
+      priceInput.dispatchEvent(new Event('change', { bubbles: true }));
+      return true;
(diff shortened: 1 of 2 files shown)


tradesdontlie:
Thanks — the stale [class*=alert] selector diagnosis was correct. Rather than chasing TradingView's obfuscated class names, #301 (merged) moved alert_create/alert_delete to the pricealerts REST API entirely, which removes DOM fragility for good. Verified live against Desktop 3.1.0.

PR #170, opened 2026-05-19
Fix tab_switch: reattach CDP client to activated tab

creativepulsenow:
## Summary

`switchTab` previously activated the requested tab in TradingView's UI but the cached CDP client remained attached to whichever target was active when CDP first connected. All subsequent `chart_get_state` / `data_get_pine_labels` / `quote_get` / etc. calls returned data from the **original** tab even though the UI showed the new one.

## Reproduction (before patch)

```js
await switchTab({ index: 2 });          // UI navigates to tab 2
const s = await chart_get_state();      // returns tab 0's symbol ❌
```

## Fix

- `connection.js` exports `setPreferredTarget(targetId)`. `findChartTarget()` prefers it on next reconnect, then falls through if stale (e.g., tab was closed).
- `tab.js` `switchTab` calls `setPreferredTarget(target.id) + disconnect()` after the `/json/activate/` call, forcing the next `evaluate()` to reconnect to the activated target.

Single-tab use cases are unaffected — `preferredTargetId` defaults to `null` and `findChartTarget` falls back to the original \"first chart target\" logic.

## Test plan

- [x] Switching across 4 chart tabs returns each chart's own symbol (downstream consumer verified end-to-end, 7/7 checks pass).
- [x] Round-trip: switch back to tab 0 returns tab 0's symbol.
- [x] `data_get_pine_labels` and `quote_get` target the active tab after switch.
- [x] Single-tab flows unchanged (no behavior change when only one chart tab exists).

### DIFF
--- a/package.json
+++ b/package.json
@@ -14,11 +14,11 @@
   "scripts": {
     "start": "node src/server.js",
     "tv": "node src/cli/index.js",
-    "test": "node --test tests/e2e.test.js tests/pine_analyze.test.js",
+    "test": "node --test tests/e2e.test.js tests/pine_analyze.test.js tests/tab.test.js",
     "test:e2e": "node --test tests/e2e.test.js",
-    "test:unit": "node --test tests/pine_analyze.test.js tests/cli.test.js",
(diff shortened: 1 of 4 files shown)


tradesdontlie:

PR #178, opened 2026-05-23
fix(win32): launch MSIX/Store TradingView via ApplicationActivationManager

tnca68-maker:
## Summary

`tv_launch` currently fails for any user who installed TradingView Desktop from the Microsoft Store / .msix. The exe lives under `C:\Program Files\WindowsApps\TradingView.Desktop_<ver>_x64__<hash>\`, which is ACL-locked — `child_process.spawn()` returns `EACCES` ("Access is denied"). The existing launcher only checks three classic exe paths and throws *TradingView not found on win32*.

This PR adds a Windows fallback that:

1. Probes for the MSIX install via `Get-AppxPackage -Name TradingView.Desktop`.
2. If found, launches it through Windows' `IApplicationActivationManager` COM interface — the **documented** way to pass arguments (here `--remote-debugging-port=9222`) to packaged apps. `Start-Process`, `spawn`, and `ShellExecute` on the WindowsApps exe all fail or strip args; `ActivateApplication` works.
3. Falls through to the existing classic-exe search if no MSIX package is found. Non-MSIX installs and non-Windows platforms are untouched.

## Verification

Tested on `TradingView.Desktop_3.1.0.7818_x64__n534cwy3pjxzj` (Microsoft Store, Electron 38 / Chrome 140):

```
{
  "success": true,
  "platform": "win32",
  "binary": "msix:TradingView.Desktop_n534cwy3pjxzj!TradingView.Desktop",
  "cdp_port": 9222,
  "browser": "Chrome/140.0.7339.133",
  "user_agent": "... TradingView/3.1.0 ... Electron/38.2.2 ..."
}
```

`tv_health_check` then returns `cdp_connected: true` with the expected `chart_symbol` and `api_available: true`.

> Note: this directly contradicts PR #114, which claims the Store sandbox *"ignores `--remote-debugging-port=9222` and does not propagate `ELECTRON_REMOTE_DEBUGGING_PORT`, so CDP port 9222 never opens."* That's not true for current 3.1.0 — the flag is honored when delivered via `ActivateApplication`. The reason direct `spawn()` fails is WindowsApps ACLs, not flag suppression. A code fix is feasible; the docs-only workaround in #114 isn't needed.

## Why this approach vs. other open MSIX PRs

There are several open PRs in this area (#52, #73, #76, #79, #93, #100, #103, #110, #135, #153, #159). Most use `Get-AppxPackage` to **locate** the exe but then still try to `spawn()` it from `WindowsApps`, which fails on a properly ACL'd Store install (or only works if the user disabled WindowsApps ACLs). `ApplicationActivationManager` is the supported path; this is the only one I've seen tested end-to-end against a current build.

## Test changes

- `tests/launch.test.js` (new): unit tests for `tryLaunchMsixWindows()` with stubbed `execSync` covering the `NOT_INSTALLED` branch, successful activation parse, PowerShell error, unparseable stdout, and the `killFirst` smoke check. **5/5 pass.**
- `tests/e2e.test.js` "tv_launch — auto-detect binary": made platform-aware. The existing test only checked macOS paths and therefore failed unconditionally on Windows and Linux on `main` (regardless of any launcher logic). Now checks the correct paths per platform and falls back to `Get-AppxPackage` on Windows.

## Refactor

The CDP-ready polling loop is now a `waitForCdp(cdpPort)` helper so the classic and MSIX paths share it instead of duplicating ~25 lines.

## Scope

Per `CONTRIBUTING.md`: this is "improving reliability of existing tools (better selectors, error handling, timeouts)." It does not connect directly to TradingView's servers, scrape data, or enable automated trading. No new runtime dependencies — only `child_process.execSync` (already imported).

## Test plan

- [x] `npm test` (`tests/pine_analyze.test.js` and unaffected suites still pass on Windows)
- [x] `node --test tests/launch.test.js` → 5/5 pass
- [x] `node --test --test-name-pattern="tv_launch" tests/e2e.test.js` → pass on Windows MSIX install
- [x] Manual end-to-end against TradingView.Desktop 3.1.0.7818 from the Microsoft Store → `tv_health_check` returns `cdp_connected: true`
- [ ] Reviewer to verify against a classic (non-MSIX) Windows install that the fallthrough still works (should be unchanged — code path untouched)

### DIFF
--- a/src/core/health.js
+++ b/src/core/health.js
@@ -2,8 +2,10 @@
  * Core health/discovery/launch logic.
  */
 import { getClient, getTargetInfo, evaluate } from '../connection.js';
-import { existsSync } from 'fs';
+import { existsSync, writeFileSync, unlinkSync } from 'fs';
 import { execSync, spawn } from 'child_process';
+import { tmpdir } from 'os';
+import { join } from 'path';
 
 export async function healthCheck() {
   await getClient();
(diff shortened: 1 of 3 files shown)


tradesdontlie:

PR #179, opened 2026-05-23
feat(batch): stable BARS_PATH for get_ohlcv + screenshot action

caakshayk1-boop:
## Summary

- **get_ohlcv**: switched to the same direct bar storage path used by `data_get_ohlcv` — eliminates Promise rejection risk that occurred when chart state wasn't fully settled
- **screenshot**: added screenshot capture action to `batch_run` with auto-created `screenshots/` dir and timestamped filenames (`batch_{symbol}_{tf}_{ts}.png`)
- `package-lock.json`: minor dependency sync

## Test plan

- [ ] Run `batch_run` with `action: "get_ohlcv"` across multiple symbols — verify no Promise rejections
- [ ] Run `batch_run` with `action: "screenshot"` — verify PNG files appear in `screenshots/`
- [ ] Run `batch_run` with unknown action — verify error returned cleanly

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -10,6 +10,9 @@
       "dependencies": {
         "@modelcontextprotocol/sdk": "^1.12.1",
         "chrome-remote-interface": "^0.33.2"
+      },
+      "bin": {
+        "tv": "src/cli/index.js"
       }
     },
     "node_modules/@hono/node-server": {
--- a/src/core/batch.js
+++ b/src/core/batch.js
@@ -1,8 +1,12 @@
 /**
  * Core batch execution logic.
  */
-import { evaluate, evaluateAsync, getClient, getChartApi, getChartCollection, safeString } from '../connection.js';
+import { evaluate, evaluateAsync, getClient, getChartApi, getChartCollection, safeString, KNOWN_PATHS } from '../connection.js';
 import { waitForChartReady } from '../wait.js';
+
+// Same path used by data_get_ohlcv — reads directly from chart's bar storage (no Promise, no rejection)
+const CHART_API  = 'window.TradingViewApi._activeChartWidgetWV.value()';
+const BARS_PATH  = KNOWN_PATHS ? KNOWN_PATHS.mainSeriesBars : `${CHART_API}._chartWidget.model().mainSeries().bars()`;
 import { writeFileSync, mkdirSync } from 'fs';
 import { join, dirname } from 'path';
 import { fileURLToPath } from 'url';
@@ -15,20 +19,15 @@ export async function batchRun({ symbols, timeframes, action, delay_ms, ohlcv_co
   const delay = delay_ms || 2000;
   const results = [];
 
-  let colPath, apiPath;
-  try { colPath = await getChartCollection(); } catch {}
-  try { apiPath = await getChartApi(); } catch {}
-
   for (const symbol of symbols) {
     for (const tf of tfs) {
       const combo = { symbol, timeframe: tf };
       try {
-        if (colPath) await evaluate(`${colPath}.setSymbol(${safeString(symbol)})`);
-        else if (apiPath) await evaluate(`${apiPath}.setSymbol(${safeString(symbol)})`);
+        // Use the same stable CHART_API path that chart_set_symbol uses
+        await evaluate(`${CHART_API}.setSymbol(${safeString(symbol)})`);
 
         if (tf) {
-          if (colPath) await evaluate(`${colPath}.setResolution(${safeString(tf)})`);
-          else if (apiPath) await evaluate(`${apiPath}.setResolution(${safeString(tf)})`);
+          await evaluate(`${CHART_API}.setResolution(${safeString(tf)})`);
         }
 
         await waitForChartReady(symbol);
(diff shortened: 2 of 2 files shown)


hsuliyin630912-code:
Independent verification that this is still needed, plus one detail that I think strengthens the case for this PR.

Still reproduces on current main
Reproduced today at c05b8f5 (fetched immediately before testing — no upstream commits since 2026-07-28), TradingView Desktop over CDP:

batch_run({
  symbols: ["OANDA:XAUUSD", "BINANCE:BTCUSDT.P"],
  action: "get_ohlcv",
  ohlcv_count: 5,
  delay_ms: 2000
})
{
  "success": true, "total_iterations": 2, "successful": 0, "failed": 2,
  "results": [
    { "symbol": "OANDA:XAUUSD",      "success": false, "error": "JS evaluation error: Uncaught (in promise)" },
    { "symbol": "BINANCE:BTCUSDT.P", "success": false, "error": "JS evaluation error: Uncaught (in promise)" }
  ]
}
Control — same symbol, same session, different action succeeds:

batch_run({ symbols: ["OANDA:XAUUSD"], action: "screenshot", delay_ms: 2000 })
// → { "successful": 1, "failed": 0 }
So symbol switching, waitForChartReady and the iteration loop are all healthy — only the get_ohlcv branch is affected. data_get_ohlcv also returns bars fine on the same chart at the same moment.

The failure looks categorical rather than timing-related
The PR description attributes this to "Promise rejection risk that occurred when chart state wasn't fully settled". From what I can measure, the failure is unconditional — exportData() is refused outright, so no amount of settling or retrying would help. Calling it directly in the page context, in the same session as the run above, and inspecting the rejection:

window.TradingViewApi._activeChartWidgetWV.value()
  .exportData({ includeTime: true, includeSeries: true, includeStudies: false })
  .catch(e => ({
    errType: Object.prototype.toString.call(e),
    isError: e instanceof Error,
    message: e && e.message ? e.message : String(e)
  }))
{
  "state": "rejected",
  "message": "Data export is not supported",
  "errType": "[object String]",
  "isError": false
}
I mention it only because it changes what the fix is: if it were a settling issue, a wait/retry would be a valid alternative — but it isn't, so switching to the direct bars path (what this PR does) is the only thing that can work. grep -rn "exportData" src/ returns exactly one hit — src/core/batch.js:51 — so this PR removes the codebase's only caller of an unsupported API.

A second defect this PR doesn't cover
Note errType above: the rejection value is a plain String, not an Error. The current .catch(reject) forwards it unchanged, so CDP can't extract a message and callers only ever see "Uncaught (in promise)" — the actual reason ("Data export is not supported") is destroyed by the error handling itself. That's why this reads like a flaky evaluation problem rather than an unsupported API.

This PR makes the message moot for get_ohlcv specifically, but the swallowing is generic to evaluateAsync callers. Might be worth normalising at the boundary:

.catch(function (e) { reject(e instanceof Error ? e : new Error(String(e))); })
Happy to open that as a separate PR if it's out of scope here.

Cross-reference
[#207](https://github.com/hsuliyin630912-code/ai-studio/issues/207) proposed the same direct-bars fix and was closed by its author; this appears to be the only open PR addressing it. Nothing in either thread mentions the "Data export is not supported" rejection, which is why I'm adding it here.

chasewhughes:
Independent verification at current main (`c05b8f5`, same as upstream HEAD at time of writing).

**Environment:** macOS 26.6 (arm64), Node v22.16.0, TradingView Desktop 3.3.0, CDP 9222.

### Reproduced the bug first, on unmodified main

```
batch_run({ symbols: ["SPCFD:SPX","TVC:DXY"], action: "get_ohlcv", timeframes: ["D"], ohlcv_count: 5 })
→ { "successful": 0, "failed": 2,
    "results": [ { "symbol": "SPCFD:SPX", "success": false,
                   "error": "JS evaluation error: Uncaught (in promise)" }, ... ] }
```

### Root cause matches this PR's diagnosis

`apiPath.exportData()` returns a Promise that rejects when the series isn't fully materialised, and the rejection surfaces through `evaluateAsync` as `Uncaught (in promise)`. `data_get_ohlcv` never hits this because it reads `KNOWN_PATHS.mainSeriesBars` synchronously.

### With this PR's `get_ohlcv` hunk applied

```
batch_run({ symbols: ["SPCFD:SPX","TVC:DXY","BATS:IWM"], action: "get_ohlcv", timeframes: ["D"], ohlcv_count: 3 })
→ { "total_iterations": 3, "successful": 3, "failed": 0 }
     SPCFD:SPX   3 bars, last close 7798.99
     TVC:DXY     3 bars, last close   99.948
     BATS:IWM    3 bars, last close  303.50
```

Closes cross-checked against `data_get_ohlcv` on the same symbols — identical.

`npm run lint` → 0 errors (4 pre-existing warnings). `npm run test:unit` → 152/152 pass.

### Two notes for whoever picks this up

1. The PR is currently `CONFLICTING` against main and needs a rebase.

2. I applied **only** the `src/core/batch.js` `get_ohlcv` hunk, and deliberately skipped this PR's two other changes:
   - the `package-lock.json` `bin` addition, which looks unrelated to the fix;
   - replacing the `colPath`/`apiPath` fallback with a hardcoded `CHART_API` for `setSymbol`/`setResolution`.

   Symbol switching was never part of this bug — it works fine on main — so folding it in widens the blast radius for no gain. A minimal diff touching only the `get_ohlcv` branch would be easier to review and merge, and CONTRIBUTING asks for the smallest equivalent change.

One follow-up worth a separate PR once this lands: batch `get_ohlcv` returns full bars, so a 20-symbol scan at the default 100 bars returns 2,000 bars in one payload. Reusing the `summary` shape from `getOhlcv()` would make batch usable for wide scans, which is its main use case.

Happy to open a rebased minimal version if the author has moved on.

PR #182, opened 2026-05-24
Fix alert_create by using internal pricealerts REST API

david1asher:
## Summary

`alert_create` has been silently broken — every call returns `success: false` with `price_set: false` and no alert is created. Root cause: the DOM-fallback in `src/core/alerts.js` queries `[class*=\"alert\"] input[type=\"text\"|number]` but TV's real alert dialog uses scoped hash classes (e.g. `submitBtn-m9pp3wEB`), so neither the price input nor the Create button is matched.

This PR swaps the DOM scrape for a direct `POST https://pricealerts.tradingview.com/create_alert` — the same endpoint TV's own dialog hits when the user clicks **Create**.

## How the payload shape was found

Hooked `window.fetch` in the page context, then clicked the native **Create alert** button in the toolbar and captured the outgoing request. The wrapped envelope and field names didn't match what the previous code was trying:

```jsonc
{
  \"payload\": {
    \"conditions\": [{                                  // plural array, not \"condition\"
      \"type\": \"cross\",
      \"frequency\": \"on_first_fire\",
      \"series\": [{ \"type\": \"barset\" }, { \"type\": \"value\", \"value\": 140.65 }],
      \"resolution\": \"1\"
    }],
    \"symbol\": \"={\\\"adjustment\\\":\\\"splits\\\",\\\"currency-id\\\":\\\"USD\\\",\\\"session\\\":\\\"extended\\\",\\\"symbol\\\":\\\"BATS:RDDT\\\"}\",
    \"resolution\": \"1\",
    \"message\": \"...\",
    \"expiration\": \"2026-06-24T18:09:21.000Z\",       // ISO 8601 string, not unix
    \"active\": true,
    \"popup\": true,
    \"mobile_push\": true,
    \"sound_file\": \"alert/fired\",
    \"sound_duration\": 0,
    \"auto_deactivate\": true,
    \"email\": false,
    \"sms_over_email\": false,
    \"web_hook\": null,
    \"name\": null,
    \"ignore_warnings\": true
  }
}
```

Two gotchas worth flagging:
- Body is **raw JSON without `Content-Type: application/json`**. Setting that header triggers a CORS preflight that fails. The browser treats the unset Content-Type as `text/plain` → simple request → no preflight.
- The `symbol` field is `=` + `JSON.stringify({...})` with `adjustment`, `currency-id`, `session`, `symbol`. Plain `\"BATS:RDDT\"` is rejected with `invalid_request`.

## Behavior changes

- `alert_create` now actually creates alerts (verified against 4 levels on `BATS:RDDT`).
- Adds optional params: `symbol`, `resolution`, `frequency`, `expiration_days`. When omitted, symbol + resolution come from the current chart via `${CHART_API}.symbol()` / `.resolution()`.
- Returns `alert_id`, `expiration_days`, and `source: 'internal_api'` on success.
- Uses `safeString` + `requireFinite` for the same injection / NaN guards already used across the codebase.
- `list()` and `deleteAlerts()` are unchanged.
- `condition` param is still accepted for backwards compat but is a no-op — the underlying API uses `type:cross` + `cross_interval` which fires on any crossing in either direction.

## Test plan

- [x] `node -e \"import('./src/core/alerts.js')...\"` — module loads, exports `create`, `list`, `deleteAlerts`
- [x] Live: created 4 alerts on RDDT (140.65 / 145 / 150 / 153), confirmed via `alert_list` with returned `alert_id` matching
- [x] Live: omitting `symbol` / `resolution` correctly picks them up from the current chart
- [ ] Existing e2e tests in `tests/e2e.test.js` still pass (the existing alert tests only assert button existence — should not be affected)

## Not in scope

- `alert_delete` for individual alerts — pricealerts service does not expose `/remove_alert` / `/delete_alert` (returns `no_such_endpoint`). Would need a separate session of fetch-hooking to find the real delete endpoint. `delete_all` via context menu is unchanged.

### DIFF
--- a/src/core/alerts.js
+++ b/src/core/alerts.js
@@ -1,81 +1,127 @@
 /**
  * Core alert logic.
+ *
+ * create() and list() use TradingView's internal pricealerts.tradingview.com REST API
+ * (same backend the desktop UI uses). Discovered by intercepting the network call from
+ * the "Create alert" dialog. The DOM-fallback path that earlier shipped here never
+ * matched the real alert dialog selectors and silently no-op'd.
  */
-import { evaluate, evaluateAsync, getClient, safeString } from '../connection.js';
+import { evaluate as _evaluate, evaluateAsync as _evaluateAsync, safeString, requireFinite } from '../connection.js';
 
-export async function create({ condition, price, message }) {
-  const opened = await evaluate(`
-    (function() {
-      var btn = document.querySelector('[aria-label="Create Alert"]')
-        || document.querySelector('[data-name="alerts"]');
-      if (btn) { btn.click(); return true; }
-      return false;
-    })()
-  `);
+const CHART_API = 'window.TradingViewApi._activeChartWidgetWV.value()';
+const CREATE_URL = 'https://pricealerts.tradingview.com/create_alert';
+const LIST_URL = 'https://pricealerts.tradingview.com/list_alerts';
 
-  if (!opened) {
-    const client = await getClient();
-    await client.Input.dispatchKeyEvent({ type: 'keyDown', modifiers: 1, key: 'a', code: 'KeyA', windowsVirtualKeyCode: 65 });
-    await client.Input.dispatchKeyEvent({ type: 'keyUp', key: 'a', code: 'KeyA' });
-  }
-
-  await new Promise(r => setTimeout(r, 1000));
-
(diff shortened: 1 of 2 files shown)


david1asher:
Closing — replacing with a cleaned-up version.

PR #183, opened 2026-05-24
Fix alert_create by using the internal pricealerts REST API

david1asher:
## What's wrong

`alert_create` is silently broken — every call returns `success: false` with `price_set: false` and no alert ever lands in TradingView. The DOM-fallback in `src/core/alerts.js` queries `[class*=\"alert\"] input[type=\"text\"|number]`, but the real alert dialog uses scoped hash classes (e.g. `submitBtn-m9pp3wEB`), so neither the price input nor the Create button get matched.

## The fix

Drop the DOM scrape and POST directly to `https://pricealerts.tradingview.com/create_alert` — the same endpoint the dialog hits when you click **Create**. I figured out the payload shape by hooking `window.fetch` in the page context and triggering the native dialog, then comparing what the previous code was sending. It wasn't close:

```jsonc
{
  \"payload\": {
    \"conditions\": [{                                  // plural array, not \"condition\"
      \"type\": \"cross\",
      \"frequency\": \"on_first_fire\",
      \"series\": [{ \"type\": \"barset\" }, { \"type\": \"value\", \"value\": 140.65 }],
      \"resolution\": \"1\"
    }],
    \"symbol\": \"={\\\"adjustment\\\":\\\"splits\\\",\\\"currency-id\\\":\\\"USD\\\",\\\"session\\\":\\\"extended\\\",\\\"symbol\\\":\\\"BATS:RDDT\\\"}\",
    \"resolution\": \"1\",
    \"message\": \"...\",
    \"expiration\": \"2026-06-24T18:09:21.000Z\",       // ISO 8601, not unix
    \"active\": true,
    \"popup\": true,
    \"mobile_push\": true,
    \"sound_file\": \"alert/fired\",
    \"sound_duration\": 0,
    \"auto_deactivate\": true,
    \"email\": false,
    \"sms_over_email\": false,
    \"web_hook\": null,
    \"name\": null,
    \"ignore_warnings\": true
  }
}
```

A couple of gotchas worth flagging if anyone else hits the same endpoint:

- The body is raw JSON **without** a `Content-Type: application/json` header. Setting that triggers a CORS preflight that fails; leaving it unset keeps the request \"simple\" and skips preflight.
- A plain `\"BATS:RDDT\"` symbol gets rejected with `invalid_request`. The `=` + JSON wrapper with `adjustment`, `currency-id`, `session`, and `symbol` is required.

## What changes

- `alert_create` actually creates alerts now.
- `symbol` and `resolution` default to whatever the chart currently shows, so existing callers don't need to change anything.
- Added optional `symbol`, `resolution`, `frequency`, and `expiration_days` params for callers that want to target another chart or override defaults.
- Returns `alert_id` from the service plus `source: 'internal_api'` on success.
- `condition` is still accepted but no longer does anything — the underlying API uses `type:cross` + `cross_interval`, which fires on any crossing in either direction. Kept the param so existing callers don't break.
- Uses the same `safeString` + `requireFinite` guards as the rest of the codebase.
- `list()` and `deleteAlerts()` are untouched.

## Tested

- `import('./src/core/alerts.js')` loads cleanly, exports `create` / `list` / `deleteAlerts`.
- Created four alerts on `BATS:RDDT` (140.65 / 145 / 150 / 153) end-to-end. Confirmed they show up in `alert_list` with the returned `alert_id`s and fire correctly.
- Omitting `symbol` / `resolution` picks them up from the active chart, as expected.
- The existing alert tests in `tests/e2e.test.js` only assert DOM button existence, so they're unaffected.

## Not in scope

I tried to also fix individual alert deletion while I was in there, but `pricealerts.tradingview.com` doesn't expose `/remove_alert`, `/delete_alert`, `/remove`, or `/delete` — all return `no_such_endpoint`. Finding the real delete endpoint needs another round of fetch-hooking and felt like its own PR. `delete_all` via the context-menu path is unchanged.

### DIFF
--- a/src/core/alerts.js
+++ b/src/core/alerts.js
@@ -1,16 +1,20 @@
 /**
  * Core alert logic.
  *
- * create() and list() use TradingView's internal pricealerts.tradingview.com REST API
- * (same backend the desktop UI uses). Discovered by intercepting the network call from
- * the "Create alert" dialog. The DOM-fallback path that earlier shipped here never
- * matched the real alert dialog selectors and silently no-op'd.
+ * create(), list(), and deleteAlerts() all hit TradingView's internal
+ * pricealerts.tradingview.com REST API — the same backend the desktop UI uses.
+ * Endpoint shapes were reverse-engineered by hooking window.fetch in the page
+ * context and triggering the native dialog / list / delete flows. The
+ * DOM-fallback path that earlier shipped here never matched the real alert
+ * dialog's selectors and silently no-op'd.
  */
 import { evaluate as _evaluate, evaluateAsync as _evaluateAsync, safeString, requireFinite } from '../connection.js';
 
 const CHART_API = 'window.TradingViewApi._activeChartWidgetWV.value()';
 const CREATE_URL = 'https://pricealerts.tradingview.com/create_alert';
 const LIST_URL = 'https://pricealerts.tradingview.com/list_alerts';
+const DELETE_URL = 'https://pricealerts.tradingview.com/delete_alerts';
+const MAX_EXPIRATION_DAYS = 60; // TradingView's hard cap; requests beyond this are rejected.
 
 function _resolve(deps) {
   return {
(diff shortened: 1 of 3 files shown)


tradesdontlie:
Thanks — your fetch-hook reverse engineering of the create_alert payload was the breakthrough for this cluster, and the payload shape you documented matches what shipped. Superseded by #301 (merged), which covers the same REST create path plus targeted alert_id deletion and CLI wiring. Verified live against Desktop 3.1.0 before merging.

PR #192, opened 2026-05-28
Add Windows AppX/Store detection to tv_launch

The-Electrician:
## Summary

- **`tv_launch` now auto-detects Microsoft Store (AppX) installs on Windows.** TradingView installed via the Microsoft Store lives under `C:\Program Files\WindowsApps` and is invisible to the existing path-based and `where` lookups. This adds a `Get-AppxPackage` fallback that resolves the install location automatically.
- **Added Quick-Connect section to CLAUDE.md** documenting the AppX launch sequence for faster session startup.

## Why

Users who install TradingView Desktop from the Microsoft Store (rather than the standalone installer) can't use `tv_launch` — it fails with "TradingView not found" because WindowsApps paths aren't in the standard search list and `where` can't find them.

## Changes

- `src/core/health.js`: Added AppX detection block between the static path search and the `where`/`which` fallback. Uses PowerShell `Get-AppxPackage` to resolve the install path.
- `CLAUDE.md`: Added Quick-Connect section with PowerShell commands for AppX launches.
- `package-lock.json`: Lockfile sync (bin entry from existing package.json).

## Test plan

- [ ] Verify `tv_launch` succeeds on a Windows machine with TradingView installed from the Microsoft Store
- [ ] Verify `tv_launch` still works on machines with the standalone installer (existing paths unchanged)
- [ ] Verify `tv_health_check` passes after AppX-detected launch

Generated with [Claude Code](https://claude.com/claude-code)

### DIFF
--- a/CLAUDE.md
+++ b/CLAUDE.md
@@ -2,6 +2,33 @@
 
 68 tools for reading and controlling a live TradingView Desktop chart via CDP (port 9222).
 
+## Quick-Connect (This Machine)
+
+TradingView Desktop is installed as a **Microsoft Store (AppX) package**.
+Node.js is at `%LOCALAPPDATA%\Programs\nodejs`.
+
+To connect in one shot, run these PowerShell commands in sequence:
+
+```powershell
+# 1. Kill any existing instance (optional)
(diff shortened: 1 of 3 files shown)


tradesdontlie:

PR #193, opened 2026-05-28
fix: restore drawing.js DI by adding _resolve() to all exports

david-fitzgerald:
## Summary

Four of the five exports in `src/core/drawing.js` throw `ReferenceError: getChartApi is not defined` at runtime. Affects `draw_list`, `draw_get_properties`, `draw_remove_one`, and `draw_clear` MCP tools — only `draw_shape` works

## Root cause

Commit f23eb1b ("Add DI and full test coverage for chart.js and drawing.js sanitization") aliased the imports as `_evaluate` / `_getChartApi` so `drawShape` could call `_resolve(_deps)` and re-bind the original names into scope. The four other exports were never updated, they still reference the bare `getChartApi` / `evaluate` identifiers, which no longer exist after the alias rename.

```js
// drawShape (works) — line 11

// listDrawings / getProperties / removeOne / clearAll (broken)
const apiPath = await getChartApi();  // ← ReferenceError
```

The regression slipped through because `sanitization.test.js` only covered `drawShape`.

## Fix

- Add `_deps` parameter and `_resolve()` call to the four broken functions, mirroring `drawShape`
- Add regression tests for each of the four, asserting they execute through to `evaluate()` with the injected `apiPath`

## Test plan

- [x] `node --test tests/sanitization.test.js` → 70/70 pass (was 66/66 before)
- [x] Reproduced original error against `main`: `draw_list` returns `{success: false, error: "getChartApi is not defined"}`
- [ ] Verify `draw_list` returns shapes after this patch (requires running TradingView desktop + MCP)

### DIFF
--- a/src/core/drawing.js
+++ b/src/core/drawing.js
@@ -44,7 +44,8 @@ export async function drawShape({ shape, point, point2, overrides: overridesRaw,
   return { success: true, shape, entity_id: result?.entity_id };
 }
 
-export async function listDrawings() {
+export async function listDrawings({ _deps } = {}) {
+  const { evaluate, getChartApi } = _resolve(_deps);
   const apiPath = await getChartApi();
   const shapes = await evaluate(`
     (function() {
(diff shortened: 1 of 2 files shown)


tradesdontlie:

PR #197, opened 2026-05-29
Fix CDP chart-target selection for TradingView Desktop (Electron)

Mojobucks:
## Problem

On TradingView **Desktop** (Electron), `findChartTarget()` often attaches to the wrong CDP target. Every `file://` page in the app matches the `/tradingview/i` fallback, so the first match is frequently the window shell, the "New tab" landing page, or a helper page (tooltip, browser-api-container, drag-service) instead of the actual chart. Once cached, the bare `'1'` liveness probe in `getClient()` keeps that wrong connection alive forever — every subsequent tool call fails with "TradingViewApi is not defined".

## Fix

- **`findChartTarget()`**: match real chart pages first (`tradingview.com/chart` for web, `/chart/` in the loaded layout URL for desktop), then fall back to TradingView pages **excluding** known non-chart shells/helpers, instead of blindly taking the first match.
- **`getClient()` liveness probe**: verify the cached connection is on an actual chart page (`window.TradingViewApi._activeChartWidgetWV` present), not just any live page. A connection to a non-chart target is closed and re-established to the chart page.
- **Plays nicely with `reconnectTo()` / `tab_switch`** (#319): an explicitly selected target is sticky — if the probe fails, we first try to reconnect to that same target, and only fall back to auto-discovery when it's gone. Network errors during the sticky lookup fall back to `connect()` so the full retry/backoff path still applies.

## Scope note

The original version of this PR also made the CDP port/host configurable via env vars; that part was since implemented upstream in #318 (more thoroughly — dual env names, exported constants, IPv4 default) and has been dropped from this PR. What remains is purely the desktop target-selection fix.

## Tests

13 unit tests for target selection, probe behavior, explicit-target stickiness, and network-error fallback (mutation-tested: removing the stickiness logic or the error handling turns them red). Full suite: 165/165 green, lint clean. Verified live against TradingView Desktop on macOS — shell/helper pages excluded, real chart page selected.

### DIFF
--- a/package.json
+++ b/package.json
@@ -17,9 +17,9 @@
     "lint": "eslint src/",
     "test": "node --test tests/e2e.test.js tests/pine_analyze.test.js",
     "test:e2e": "node --test tests/e2e.test.js",
-    "test:unit": "node --test tests/pine_analyze.test.js tests/cli.test.js tests/sanitization.test.js tests/replay.test.js tests/launch.test.js tests/chart_indicator.test.js tests/chart_history.test.js tests/update.test.js",
+    "test:unit": "node --test tests/pine_analyze.test.js tests/cli.test.js tests/sanitization.test.js tests/replay.test.js tests/launch.test.js tests/chart_indicator.test.js tests/chart_history.test.js tests/update.test.js tests/connection.test.js",
     "test:cli": "node --test tests/cli.test.js",
-    "test:all": "node --test tests/e2e.test.js tests/pine_analyze.test.js tests/cli.test.js tests/sanitization.test.js tests/replay.test.js tests/launch.test.js tests/chart_indicator.test.js tests/chart_history.test.js tests/update.test.js",
+    "test:all": "node --test tests/e2e.test.js tests/pine_analyze.test.js tests/cli.test.js tests/sanitization.test.js tests/replay.test.js tests/launch.test.js tests/chart_indicator.test.js tests/chart_history.test.js tests/update.test.js tests/connection.test.js",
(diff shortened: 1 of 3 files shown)


Danielworkapp:
Reproduced on `main` and verified this PR on TradingView Desktop 3.4.0 (Microsoft Store build, Windows 11 Pro 10.0.26200, Node 24.19.0).

**Test:** kill TradingView, relaunch it with `--remote-debugging-port=9222`, call `getClient()` as soon as the first TradingView page appears in `/json/list` (before the chart page is listed), then call it again 5s after the chart page appears. Each call reports the `location.href` of the page it's attached to.

**`main` (c05b8f5):** stuck on a helper page for good, as described.

```
call 1, during load : app/browser-api-container/index.html  (TradingViewApi: undefined)
call 2, chart loaded: app/browser-api-container/index.html  (TradingViewApi: undefined)
```

**This PR (813fb08):** the new liveness probe drops the wrong page and reconnects to the chart.

```
call 1, during load : app/splash/index.html  (TradingViewApi: undefined)
call 2, chart loaded: https://www.tradingview.com/chart/<id>/  (TradingViewApi: object)
```

One gap: on 3.4.0 the first call landed on `app/splash/index.html`, which isn't in `NON_CHART`. Adding `splash` would stop the fallback from attaching to it at all. For reference, these are the non-chart pages I saw on 3.4.0, all under `resources/app.asar/app/`: `window/index.html`, `tooltip/index.html`, `browser-api-container/index.html`, `renderer-services/drag-service/index.html`, and `splash/index.html` during startup.

`npm run lint` is clean. `npm run test:unit` on Windows: 136/138, and the same two tests fail on `main` (Pine compile and the source audit). The 13 new connection tests pass.

PR #200, opened 2026-05-31
fix: evaluate ReferenceError in 3 chart tools + quote_get mislabeled data

8Link21:
## Summary

Two independent runtime bugs in the core layer.

### 1. `evaluate is not defined` in three chart functions

`src/core/chart.js` imports the CDP helper aliased — `import { evaluate as _evaluate, ... }` — so functions are expected to obtain it via `const { evaluate } = _resolve(_deps)`. Three functions skipped that line and called `evaluate(...)` directly, throwing `ReferenceError: evaluate is not defined` at runtime:

- `getVisibleRange`
- `scrollToDate` (tool `chart_scroll_to_date`)
- `symbolInfo` (tool `symbol_info`)

Fix: add the `_resolve(_deps)` destructure (and the `_deps` param) to each, matching the pattern already used by `setSymbol`, `setVisibleRange`, etc.

### 2. `quote_get` returned mislabeled data

`getQuote` in `src/core/data.js` always reads the **active chart's** bars and `symbolExt()` metadata, but labeled the result with the **requested** `symbol` argument. Calling `quote_get` with a symbol different from the chart returned the chart's OHLC under the requested symbol's name — silently wrong data.

Fix: always read the chart's true symbol; if the requested symbol differs from the chart symbol, throw an explicit error telling the caller to `chart_set_symbol` first. `quote_get` reflects the active chart only.

## Reproduction (before)

- `symbol_info` / `chart_scroll_to_date` → `{ "success": false, "error": "evaluate is not defined" }`
- Chart on symbol A; `quote_get` with symbol B → returns A's OHLC labeled as B.

## Testing

- `npm run test:unit` and `node --test tests/sanitization.test.js` pass (95 tests). The `getQuote` change also drops a user-input interpolation from the evaluated string; the sanitization audit still passes.
- Verified live over CDP: the three tools return data; cross-symbol `quote_get` now errors explicitly.

### DIFF
--- a/src/core/batch.js
+++ b/src/core/batch.js
@@ -1,8 +1,9 @@
 /**
  * Core batch execution logic.
  */
-import { evaluate, evaluateAsync, getClient, getChartApi, getChartCollection, safeString } from '../connection.js';
+import { evaluate, getClient, getChartApi, getChartCollection, safeString } from '../connection.js';
 import { waitForChartReady } from '../wait.js';
+import { getOhlcv, getPineTables } from './data.js';
 import { writeFileSync, mkdirSync } from 'fs';
 import { join, dirname } from 'path';
 import { fileURLToPath } from 'url';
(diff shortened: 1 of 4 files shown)


wiivoo:
+1 to merge. Reproduced the same `ReferenceError: evaluate is not defined` on current `main` when calling `symbol_info`, `chart_get_visible_range`, and `chart_scroll_to_date` — those three handlers in `src/core/chart.js` reference `evaluate` without resolving it via the DI helper that every working sibling (e.g. `getState`) uses. This PR's `const { evaluate } = _resolve(_deps);` fix matches the root cause exactly and resolves the error for all three tools. Thanks for the fix!

tradesdontlie:

PR #201, opened 2026-05-31
feat(windows): support TradingView installed as Windows Store (MSIX) app

sbachiraoun-ops:
## Summary

- Detect TradingView installed as an MSIX/Store package via PowerShell `Get-AppxPackage`
- Launch Store apps with `--remote-debugging-port` using `ApplicationActivationManager` COM interface (standard Win32 API for launching packaged apps with arguments)
- Fall back to `explorer.exe shell:AppsFolder` activation if COM launch fails
- Add example crypto swing-trading `rules.json` configuration
- Add BTC DCA Drop strategy (Pine Script v6) for 4H BTCUSDT

## Why

TradingView distributed via tvd-packages.tradingview.com installs as an MSIX package under `C:\Program Files\WindowsApps\`. The existing `tv_launch` tool only checked `%LOCALAPPDATA%\TradingView\` and `Program Files`, so it silently failed for Store installs. MSIX apps cannot be `spawn()`-ed directly — they require Windows COM activation APIs.

## Test plan

- [ ] TradingView installed via tvd-packages.tradingview.com (MSIX)
- [ ] `tv_launch` detects and launches the Store app with CDP port active
- [ ] `tv_health_check` returns `cdp_connected: true` after launch
- [ ] Falls back gracefully on non-Windows or non-Store installs

### DIFF
--- a/btc_dca_strategy.pine
+++ b/btc_dca_strategy.pine
@@ -1,0 +1,92 @@
+//@version=6
+strategy(title="BTC DCA Drop v7", shorttitle="BTC-DCA-v7", overlay=true, initial_capital=1000, currency=currency.USD, default_qty_type=strategy.fixed, default_qty_value=1, commission_type=strategy.commission.percent, commission_value=0.1, pyramiding=2, calc_on_every_tick=true)
+
+// ─── INPUTS ────────────────────────────────────────────────────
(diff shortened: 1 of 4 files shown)


tradesdontlie:

PR #205, opened 2026-06-02
fix: finish DI refactor — 3 chart tools throw 'evaluate is not defined' + add no-undef lint guard

xxhanselxx-sudo:
## Problem

Several MCP tools throw `evaluate is not defined` at runtime: `scroll_to_date`, `symbol_info`, `get_visible_range` (and the drawing tools, previously).

**Root cause:** the dependency-injection refactor renamed the imported `evaluate`/`getChartApi` to `_evaluate`/`_getChartApi` and routes them through a per-function `const { evaluate } = _resolve(_deps)`. A handful of functions were missed — they still call `evaluate(...)` bare, which is no longer in scope. The code parses fine and only throws when the tool is invoked.

## Fix

- `src/core/chart.js`: add the missing `const { evaluate } = _resolve(_deps)` (+ `_deps` param) to `getVisibleRange`, `scrollToDate`, `symbolInfo`.
- `src/core/drawing.js`: use the renamed `_evaluate`/`_getChartApi` in `list`/`remove`/`clear`/`getProperties`.

## Regression guard

Adds a minimal ESLint flat config (`no-undef` + a few correctness rules) and an `npm run lint` script. `no-undef` statically catches exactly this "renamed-but-not-fully-updated import" class of bug, so it can't be merged unnoticed again. `src/` is currently clean (0 errors; a few harmless unused-import warnings remain).

Suggested CI workflow (I couldn't include the file itself — the token used lacks the `workflow` scope):

```yaml
name: lint
on:
  push: { branches: [main] }
  pull_request:
jobs:
  eslint:
    runs-on: ubuntu-latest
    steps:
      - uses: actions/checkout@v4
      - uses: actions/setup-node@v4
        with: { node-version: '20' }
      - run: npm install
      - run: npm run lint
```

## Verification

- `npm run lint` -> 0 errors.
- The three tools return data instead of throwing (tested live against TradingView Desktop via CDP).

### DIFF
--- a/eslint.config.mjs
+++ b/eslint.config.mjs
@@ -1,0 +1,34 @@
+// Minimal lint guard.
+//
+// Primary purpose: catch `no-undef` ("X is not defined") — the exact class of bug
+// that an unfinished refactor introduces silently. When imports are renamed
+// (e.g. `evaluate` -> `_evaluate` behind a `_resolve(_deps)` helper) but a few
+// call sites are missed, the code parses fine and only throws at runtime.
+// `no-undef` flags those statically, so CI blocks the regression at PR time.
+//
+// Globals below are the runtime APIs used across src/ (Node + browser/CDP context).
+export default [
+  {
+    files: ['**/*.js'],
+    languageOptions: {
+      ecmaVersion: 'latest',
+      sourceType: 'module',
+      globals: {
+        fetch: 'readonly', setTimeout: 'readonly', clearTimeout: 'readonly',
+        setInterval: 'readonly', clearInterval: 'readonly', console: 'readonly',
+        process: 'readonly', Buffer: 'readonly', URL: 'readonly',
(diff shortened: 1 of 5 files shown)


tradesdontlie:
Verified live before merging: all 7 previously-broken tools (symbol_info, chart_get_visible_range, chart_scroll_to_date, draw_list, draw_get_properties, draw_remove_one, draw_clear) exercised against a running TradingView Desktop 3.1.0 via the MCP stdio protocol — all pass, including a full draw→inspect→remove round-trip. Lint guard confirmed to catch the original bug class (0 errors on the fixed tree). 109/109 unit tests pass. This was one of ~20 PRs addressing this regression — merging this one because it pairs the fix with the no-undef guard that prevents it from happening again. Thanks @xxhanselxx-sudo!

PR #214, opened 2026-06-05
Add Streamable HTTP transport alongside stdio

ramakrishna0908:
## Summary

- Extracts `McpServer` construction into `src/create-server.js` so both transports share the same tool registration
- `src/server.js` keeps the existing stdio path unchanged (no breaking change to existing Claude Code config)
- `src/server-http.js` adds a session-aware [MCP Streamable HTTP](https://modelcontextprotocol.io/specification/2025-03-26/basic/transports#streamable-http) server; sessions tracked by `Mcp-Session-Id` header, each session gets its own `McpServer` instance
- Port defaults to `3000`, configurable via `MCP_HTTP_PORT` env var or first CLI arg
- Adds `start:http` npm script

## Usage

```bash
npm run start:http           # http://127.0.0.1:3000/mcp
MCP_HTTP_PORT=8080 npm run start:http
```

Configure any MCP client that supports HTTP transport:
```json
{ "url": "http://127.0.0.1:3000/mcp" }
```

## Test plan

- [x] `npm run start` (stdio) still works
- [x] `npm run start:http` starts and responds `200 OK` to a well-formed `initialize` POST
- [x] Session ID returned in `Mcp-Session-Id` response header and usable in subsequent requests

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -10,12 +10,15 @@
       "dependencies": {
         "@modelcontextprotocol/sdk": "^1.12.1",
         "chrome-remote-interface": "^0.33.2"
+      },
+      "bin": {
+        "tv": "src/cli/index.js"
       }
     },
     "node_modules/@hono/node-server": {
-      "version": "1.19.11",
-      "resolved": "https://registry.npmjs.org/@hono/node-server/-/node-server-1.19.11.tgz",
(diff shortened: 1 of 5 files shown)


ramakrishna0908:
Closing — will keep using stdio transport for now.

PR #216, opened 2026-06-06
Fix strategy detection + getTrades mapping (data.js) and Pine add-to-chart button (locale-safe)

melodjk:
## What

Three fixes to `src/core/data.js` and `src/core/pine.js`, with test coverage:

1. **fix(data): detect strategy by reportData/ordersData, read new-tester metrics** — strategy detection now keys off `reportData`/`ordersData` and reads metrics from TradingView's newer strategy-tester UI (the prior path missed the new tester layout).
2. **feat(data): map getTrades order legs to readable keys** — `getTrades` order legs mapped to human-readable keys instead of raw indices.
3. **fix(pine): match add-to-chart button by title/aria, not English text only** — the "add to chart" button matcher uses `title`/`aria-label`, so it works on non-English TradingView locales (previously matched English button text only).

## Tests
Adds `tests/strategy_results.test.js` and `tests/pine_compile_button.test.js`.

## Notes
Branched from `main`; 3 commits ahead, 0 behind (clean fast-forward).

### DIFF
--- a/package.json
+++ b/package.json
@@ -14,9 +14,9 @@
   "scripts": {
     "start": "node src/server.js",
     "tv": "node src/cli/index.js",
-    "test": "node --test tests/e2e.test.js tests/pine_analyze.test.js",
+    "test": "node --test tests/e2e.test.js tests/pine_analyze.test.js tests/strategy_results.test.js",
     "test:e2e": "node --test tests/e2e.test.js",
-    "test:unit": "node --test tests/pine_analyze.test.js tests/cli.test.js",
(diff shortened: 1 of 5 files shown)


tradesdontlie:
The strategy detection + getTrades mapping half of this is now fixed on main by #323 (verified live — correct detection, real metrics, panel auto-open). Leaving this open for the *locale-safe Pine add-to-chart button* part, which is separate; if you want, rebase to just that piece and I'll review it.

Issue #165, opened 2026-05-18
core/tab.js ignores CDP_HOST and CDP_PORT env vars — inconsistent with connection.js


bavariacoin:
## Summary

`src/core/tab.js` hardcodes `CDP_HOST = 'localhost'` and `CDP_PORT = 9222`,
ignoring the `CDP_HOST` / `CDP_PORT` env vars that `src/connection.js`
already respects. As a result, `tab_list` and `tab_switch` always target
the MCP-server-local loopback regardless of configuration. Setups where
TradingView Desktop runs on a different host than the MCP server (e.g.
TV Desktop on Windows + MCP server in WSL2 with `CDP_HOST=<windows-host-ip>`)
cannot use these two tools.

## Reproduction

1. Launch TV Desktop on host **A** with `--remote-debugging-port=9222 --remote-debugging-address=0.0.0.0`.
2. Configure the MCP server on host **B** with env `CDP_HOST=<A-ip>`, `CDP_PORT=9222`.
3. Verify CDP is reachable from B: `curl http://<A-ip>:9222/json/version` returns the expected JSON.
4. Invoke `tab_list` via your MCP client.

**Expected:** JSON array of open TradingView chart tabs.

**Actual:** `fetch failed` (Node `fetch` rejecting on connection refused
to `localhost:9222` on host **B**, where nothing is listening).

The same failure mode applies to `tab_switch` (line 100, `/json/activate/:id`).

## Root cause

`src/core/tab.js` lines 7–8:

```js
const CDP_HOST = 'localhost';
const CDP_PORT = 9222;
```

Compared to `src/connection.js` lines 5–6, which already follows the
env-var-aware pattern:

```js
const CDP_HOST = process.env.CDP_HOST || 'localhost';
const CDP_PORT = parseInt(process.env.CDP_PORT || '9222', 10);
```

This is an inconsistency between modules, not a new feature request.
Both `fetch()` call sites in `core/tab.js` (line 14 `/json/list` and
line 100 `/json/activate/:id`) interpolate the module-level constants,
so a single fix at lines 7–8 corrects both tools.

## Proposed patch

```diff
--- a/src/core/tab.js
+++ b/src/core/tab.js
@@ -4,8 +4,8 @@
  */
 import { getClient, evaluate } from '../connection.js';

-const CDP_HOST = 'localhost';
-const CDP_PORT = 9222;
+const CDP_HOST = process.env.CDP_HOST || 'localhost';
[...]

PR #217, opened 2026-06-06
Honor TV_CDP_PORT env var instead of hardcoding the CDP port

rufusalester:
## What

Read the CDP remote-debugging port from a `TV_CDP_PORT` environment variable across the connection, tab, and health modules, falling back to `9222` when the variable is unset.

## Why

The port was hardcoded to `9222` in several places (`src/connection.js`, `src/core/tab.js`, `src/core/health.js`). On machines where `9222` is already occupied by another Chromium/WebView2-based app, there was no way to point the MCP at a different debug port without editing source — and those edits get clobbered on every `git pull`.

## Changes

- `src/connection.js`, `src/core/tab.js` — `CDP_PORT` now reads `process.env.TV_CDP_PORT ? Number(...) : 9222`.
- `src/core/health.js` — `launch()` fallback uses the same env-var-then-9222 logic.
- `src/cli/commands/health.js`, `src/tools/health.js`, `src/core/stream.js` — help/banner text mentions the `TV_CDP_PORT` override.

## Compatibility

Default behavior is unchanged: with `TV_CDP_PORT` unset, the port is still `9222`. The explicit `port`/`--port` argument still takes precedence over the env var.

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -10,6 +10,9 @@
       "dependencies": {
         "@modelcontextprotocol/sdk": "^1.12.1",
         "chrome-remote-interface": "^0.33.2"
+      },
+      "bin": {
+        "tv": "src/cli/index.js"
       }
     },
     "node_modules/@hono/node-server": {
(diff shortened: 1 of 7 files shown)


tradesdontlie:

PR #218, opened 2026-06-06
Fix 'evaluate is not defined' in 3 chart.js functions missed by the DI refactor

igorlino:
## Problem

The f23eb1b DI refactor ("Add DI and full test coverage for chart.js…") destructured `evaluate` per-function via `const { evaluate } = _resolve(_deps);` in 5 chart.js functions but skipped **3**:

- `getVisibleRange` (src/core/chart.js:118)
- `scrollToDate` (src/core/chart.js:160)
- `symbolInfo` (src/core/chart.js:199)

Each still referenced a bare `evaluate` identifier, which resolves nowhere (the module imports it as `_evaluate`) — so **every invocation of the `chart_scroll_to_date`, `symbol_info`, and `chart_get_visible_range` MCP tools (and the `scroll`/`info`/`range` CLI commands) threw `ReferenceError: evaluate is not defined`**.

## Fix

Add the `_deps` parameter + resolve line to all 3, matching their 5 siblings; `= {}` defaults on `getVisibleRange`/`symbolInfo` since the MCP tool + CLI layers call them with no arguments.

## Tests

New `tests/chart.test.js` — 5 DI tests using the `mockEvaluate`/`mockDeps` pattern from `replay.test.js`: range pass-through, scroll window math (daily + intraday), invalid-date rejection before any CDP call, symbolExt spread. Each fails with the pre-fix `ReferenceError` by construction (exercising the functions requires the resolve line).

- `node --test tests/chart.test.js` → 5/5 pass
- `npm run test:unit` → 29/29 pass
- `node --test tests/replay.test.js tests/sanitization.test.js` → 105/105 pass

## Live verification

Against a running headless container (CDP): `tv info` → full symbolExt readback, `tv range` → real visible range, `tv scroll 2026-01-15` → centered with the correct ±25-bar window. All three previously threw the ReferenceError on this build.

### DIFF
--- a/src/core/chart.js
+++ b/src/core/chart.js
@@ -115,7 +115,8 @@ export async function manageIndicator({ action, indicator, entity_id, inputs: in
   }
 }
 
-export async function getVisibleRange() {
+export async function getVisibleRange({ _deps } = {}) {
+  const { evaluate } = _resolve(_deps);
   const result = await evaluate(`
     (function() {
       var chart = ${CHART_API};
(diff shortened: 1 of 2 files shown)


tradesdontlie:

PR #222, opened 2026-06-06
fix(core): bind evaluate/getChartApi via _resolve in 7 drawing/chart fns (ReferenceError)

mirkenstein:
## Summary
Seven MCP tools throw a `ReferenceError` at runtime and return `{ "success": false, "error": "getChartApi is not defined" }` (or `"evaluate is not defined"`):

`draw_list`, `draw_remove_one`, `draw_get_properties`, `draw_clear`, `chart_scroll_to_date`, `chart_get_visible_range`, `symbol_info`.

(`draw_shape`, `chart_set_visible_range`, `chart_set_symbol`, etc. were unaffected.)

## Root cause
`src/core/drawing.js` and `src/core/chart.js` import the helpers under aliases (`_evaluate` / `_getChartApi` / `_evaluateAsync`) and expose them via `_resolve(_deps)`. `drawShape()` and `setVisibleRange()` use that pattern, but the seven functions above still call the **bare** `evaluate` / `getChartApi`, which aren't in scope → `ReferenceError`. It reproduces in both local and `TV_CDP_HOST` networked modes (not CDP-specific).

## Fix
Adopt the same `_deps` / `_resolve` pattern in each affected function:

```js
const { evaluate, getChartApi } = _resolve(_deps);  // drawing.js
const { evaluate } = _resolve(_deps);               // chart.js
```

14 insertions / 7 deletions across the two files. No behavior change beyond making the helpers resolvable; also threads an optional `_deps` for DI/testing parity with the rest of the module.

## Testing
- `node --check` on both files.
- Mock-`_deps` smoke test: all 7 functions run with **0 ReferenceErrors**.
- Verified live against TradingView through the MCP: `draw_shape` → `draw_list` (lists it) → `draw_remove_one` (removes it) → `draw_list` (confirms); `chart_scroll_to_date`, `chart_get_visible_range`, `symbol_info` all return data.

## Suggested follow-up
A unit test passing a mock `_deps` per function would catch this class of regression.

### DIFF
--- a/src/core/chart.js
+++ b/src/core/chart.js
@@ -115,7 +115,8 @@ export async function manageIndicator({ action, indicator, entity_id, inputs: in
   }
 }
 
-export async function getVisibleRange() {
+export async function getVisibleRange({ _deps } = {}) {
+  const { evaluate } = _resolve(_deps);
   const result = await evaluate(`
     (function() {
       var chart = ${CHART_API};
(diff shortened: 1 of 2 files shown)


tradesdontlie:

PR #224, opened 2026-06-07
feat(chart): load history in setVisibleRange so multi-year ranges work

mirkenstein:
## Summary
`setVisibleRange` only zoomed within **already-loaded** bars. The chart lazy-loads ~300 bars initially, so requesting a multi-year window (common at 1D) clamped the view to ~14 months — there was no way to reproduce a multi-year daily chart programmatically.

## Fix
Before zooming, page history back via `mainSeries().requestMoreData(1000)` until the earliest loaded bar reaches `from` (or `requestMoreDataAvailable()` is false, or a 25-iteration guard trips), then `zoomToBarsRange` as before. Isolated to `setVisibleRange` in `src/core/chart.js`.

## Notes
- `requestMoreData` is async (network fetch), so we poll the earliest loaded bar between requests (~1.8 s spacing) rather than relying on a return value.
- The public `chart.setVisibleRange(...)` is **"Not implemented"** on this TradingView Desktop build, hence the internal `requestMoreData` approach.

## Testing
- `node --check`.
- Mock-deps: returns immediately (no loop hang).
- **Verified live** on `OANDA:XAUUSD` (1D): requesting `from` 2019-06 → `to` 2026-09 grew the loaded history from ~300 bars (first 2021-05) to **2,300 bars (back to 2017-07)** and zoomed the visible range to **2019-06 → present**.

### DIFF
--- a/src/core/chart.js
+++ b/src/core/chart.js
@@ -129,6 +129,23 @@ export async function setVisibleRange({ from, to, _deps }) {
   const { evaluate } = _resolve(_deps);
   const f = requireFinite(from, 'from');
   const t = requireFinite(to, 'to');
+
+  // Ensure enough history is loaded to cover `from`. The chart lazy-loads bars
+  // (~300 initially), so without this a multi-year range clamps to whatever is
+  // already loaded. Page back via requestMoreData until the earliest loaded bar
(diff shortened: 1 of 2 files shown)


jitendrasinghofficial17-stack (commented):
Awesome

PR #227, opened 2026-06-07
fix(chart): pass study inputs as an object so indicator params apply

mirkenstein:
## Summary
`chart_manage_indicator` (add) built the `createStudy` inputs as an array of `{ id, value }` objects. The charting library **silently ignores** that shape, so every study fell back to its **default inputs**.

Most visible symptom: adding a `Moving Average` with `{ length: 50 }` and another with `{ length: 200 }` produced **two identical MA(9) lines** (same legend value, one visible curve) instead of a 50 and a 200.

## Fix
Pass `inputs` as the plain `{ inputId: value }` object `createStudy` actually expects. Isolated to the `add` branch of `manageIndicator` in `src/core/chart.js`.

```diff
-    const inputArr = inputs ? Object.entries(inputs).map(([k, v]) => ({ id: k, value: v })) : [];
+    const inputObj = inputs && typeof inputs === 'object' ? inputs : {};
...
-        chart.createStudy(${safeString(indicator)}, false, false, ${JSON.stringify(inputArr)});
+        chart.createStudy(${safeString(indicator)}, false, false, ${JSON.stringify(inputObj)});
```

The tool already accepts `inputs` as a JSON string and parses it to an object upstream, so the public tool interface is unchanged.

## Testing
- `node --check src/core/chart.js`.
- **Verified live** on `OANDA:XAUUSD` (1D): before the fix, two "Moving Average" adds with `{length:50}` / `{length:200}` both read `MA 4,467.846` (length 9). After the fix, the legend reads `MA 50 close 4,627.385` and `MA 200 close 4,427.760` with two distinct curves. Confirmed via `getStudyById(id).getInputValues()` returning `length: 50` / `length: 200`.
- Cross-checked the format empirically: object `{length:200}` and positional `[200]` both apply; the prior `[{id,value}]` form does not.

### DIFF
--- a/src/core/chart.js
+++ b/src/core/chart.js
@@ -89,12 +89,16 @@ export async function manageIndicator({ action, indicator, entity_id, inputs: in
   const inputs = inputsRaw ? (typeof inputsRaw === 'string' ? JSON.parse(inputsRaw) : inputsRaw) : undefined;
 
   if (action === 'add') {
-    const inputArr = inputs ? Object.entries(inputs).map(([k, v]) => ({ id: k, value: v })) : [];
+    // createStudy expects inputs as a plain { inputId: value } object. The
+    // [{ id, value }] form is silently ignored by the charting library, leaving
+    // the study at its default inputs — e.g. a "Moving Average" requested with
+    // { length: 200 } would stay at the default length (9). Pass the object map.
+    const inputObj = inputs && typeof inputs === 'object' ? inputs : {};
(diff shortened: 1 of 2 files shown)


tradesdontlie:
Thanks — correct diagnosis of the inputs bug. Superseded by #322 (merged), which fixes it by applying inputs post-create via the study's getInputValues/setInputValues and reading back to confirm (createStudy's 4th arg is unreliable across builds). Verified live: EMA with {length:99} now reads back 99. Fixes #249.

PR #229, opened 2026-06-07
fix: wire sanitization and replay tests into test:unit script

mikebikes:
## Summary

- `tests/sanitization.test.js` (36 tests) and `tests/replay.test.js` (21 tests) existed in the repo but were never included in any npm script — they would silently not run for contributors
- Added both files to `test:unit` and `test:all` so all offline tests are discoverable and run together
- Updated stale test count in README and CONTRIBUTING (29 → 86)
- Fixed CONTRIBUTING to reference `npm run test:unit` instead of `npm test`, which runs e2e tests that require a live TradingView instance

## Test plan

- [ ] `npm run test:unit` should now run all 86 offline tests (pine_analyze, cli, sanitization, replay)
- [ ] `npm run test:all` covers all offline + e2e tests
- [ ] No logic changes — docs and package.json scripts only

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -32,13 +32,13 @@ If you're unsure whether a feature fits, open an issue to discuss before submitt
 
 ```bash
 npm install
-npm test          # 29 offline tests (no TradingView needed)
-tv status         # verify CDP connection (TradingView must be running)
+npm run test:unit  # 86 offline tests (no TradingView needed)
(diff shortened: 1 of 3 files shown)


tradesdontlie:
Thanks — this was a real gap. Superseded by #317, which wired tests/sanitization.test.js, tests/replay.test.js, and the new tests/launch.test.js into test:unit and test:all (109 tests now run and pass). Closing.

PR #230, opened 2026-06-07
Fix drawing tools throwing "getChartApi is not defined" + Windows tes…

hamad26650:
…t path

src/core/drawing.js: the _resolve(_deps) refactor only wired drawShape(). listDrawings/getProperties/removeOne/clearAll still referenced bare `getChartApi`/`evaluate`, which are imported under aliased names (_getChartApi/_evaluate) — so every call threw a ReferenceError at runtime (draw_clear, draw_list, draw_remove_one, draw_get_properties were all broken). Bring evaluate/getChartApi into local scope via _resolve(_deps) in all four, matching drawShape.

tests/drawing.test.js: new regression coverage for the five drawing functions using the same DI/mock pattern as replay.test.js.

tests/sanitization.test.js: the source-audit describe used `new URL(...).pathname`, which yields a leading-slash path ("/C:/...") that readdirSync mangles to "C:\C:\..." on Windows. Use fileURLToPath() so the suite runs on Windows.

### DIFF
--- a/src/core/drawing.js
+++ b/src/core/drawing.js
@@ -44,7 +44,8 @@ export async function drawShape({ shape, point, point2, overrides: overridesRaw,
   return { success: true, shape, entity_id: result?.entity_id };
 }
 
-export async function listDrawings() {
+export async function listDrawings({ _deps } = {}) {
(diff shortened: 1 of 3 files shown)


tradesdontlie:

PR #237, opened 2026-06-09
Fix CDP connection and launcher for Windows Store installs

Mukul-kepl26:
## Summary

Two Windows-specific fixes that prevented the MCP server from connecting to TradingView Desktop:

### 1. Use IPv4 loopback instead of `localhost` for CDP
`CDP_HOST` is now `127.0.0.1` (was `localhost`), and the launcher's readiness poll uses `127.0.0.1` too. Chrome/Electron's `--remote-debugging-port` binds **only** to `127.0.0.1`, but on Windows `localhost` frequently resolves to IPv6 `::1` first — which the debug server does not listen on — causing connection failures even when TradingView is running with CDP enabled.

### 2. Detect Microsoft Store (Appx) installs in `launch()`
TradingView is commonly installed under `C:\Program Files\WindowsApps\<pkg>`, a path the static candidate list never covers. `launch()` now falls back to `Get-AppxPackage -Name '*TradingView*'` to resolve the install location, so `tv_launch` works on Store installs.

## Testing
- Verified `tv_health_check` connects to a live chart after these changes (CDP on `127.0.0.1:9222`).
- Both files pass `node --check`.

### DIFF
--- a/src/connection.js
+++ b/src/connection.js
@@ -2,7 +2,10 @@ import CDP from 'chrome-remote-interface';
 
 let client = null;
 let targetInfo = null;
-const CDP_HOST = 'localhost';
+// Use the IPv4 loopback explicitly. Chrome/Electron's --remote-debugging-port
+// binds only to 127.0.0.1; on Windows 'localhost' often resolves to IPv6 ::1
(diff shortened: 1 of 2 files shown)


tradesdontlie:

PR #239, opened 2026-06-10
chore: sync package-lock.json with tv bin entry

Bachu-28:
## What changed
Adds the missing `bin` entry (`tv` → `src/cli/index.js`) to `package-lock.json`, syncing the lockfile with what `package.json` already declares.

## Why
Running a fresh `npm install` reconciles the lockfile and produces this 3-line diff. Committing it keeps installs reproducible and avoids an unexpected dirty working tree for the next person who installs dependencies.

## Notes for reviewer
- No source changes — lockfile only.
- Diff is 3 insertions; no dependency versions changed.

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -10,6 +10,9 @@
       "dependencies": {
         "@modelcontextprotocol/sdk": "^1.12.1",
         "chrome-remote-interface": "^0.33.2"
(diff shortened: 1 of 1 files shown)


tradesdontlie:

PR #245, opened 2026-06-12
Add Microsoft Store (MSIX) detection to TradingView launch on Windows

maxvama-beep:
## Problem

`tv_launch` (MCP tool) and `tv launch` (CLI) fail to find TradingView on Windows when it is installed from the Microsoft Store (MSIX package). The search only covered `%LOCALAPPDATA%\TradingView`, `Program Files`, and `Program Files (x86)`, and `where TradingView.exe` does not find Store installs either.

## Changes

- **Extracted path resolution** from `launch()` into an exported `findTradingViewPath({ platform, env, _deps })` in `src/core/health.js`, following the repo's DI pattern so it is unit-testable. Both the CLI and MCP tool route through `core.launch()`, so one fix covers both.
- **MSIX detection on win32** after the static paths miss:
  1. `Get-AppxPackage -Name TradingView.Desktop` via PowerShell (primary — the `WindowsApps` directory is ACL-restricted on most machines)
  2. Directory glob of `C:\Program Files\WindowsApps\TradingView.Desktop_*_x64__*\TradingView.exe` as fallback, picking the highest version numerically when multiple packages exist
- The "TradingView not found" error now lists the MSIX location among searched paths.
- **Tests**: new `tests/launch.test.js` with 9 unit tests (standard installs, both MSIX routes, version ordering, non-x64 filtering, ACL-denied directory, macOS/Linux regressions). Fixed the e2e `tv_launch` path test which was hardcoded to macOS-only paths and failed on every Windows machine. Registered `launch.test.js` and the previously unregistered `replay.test.js` in the npm test scripts.

## Verification

- All 9 new unit tests pass; replay unit tests still pass (48/48 combined).
- Verified live on a machine with the Store install: resolution finds `C:\Program Files\WindowsApps\TradingView.Desktop_3.2.0.7916_x64__n534cwy3pjxzj\TradingView.exe`, and launching that exe with `--remote-debugging-port=9222` works.
- e2e `tv_launch` test passes against the live app.

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -10,6 +10,9 @@
       "dependencies": {
         "@modelcontextprotocol/sdk": "^1.12.1",
         "chrome-remote-interface": "^0.33.2"
+      },
+      "bin": {
+        "tv": "src/cli/index.js"
       }
     },
     "node_modules/@hono/node-server": {
--- a/package.json
+++ b/package.json
@@ -16,9 +16,9 @@
     "tv": "node src/cli/index.js",
     "test": "node --test tests/e2e.test.js tests/pine_analyze.test.js tests/launch.test.js",
     "test:e2e": "node --test tests/e2e.test.js",
-    "test:unit": "node --test tests/pine_analyze.test.js tests/cli.test.js tests/launch.test.js tests/replay.test.js",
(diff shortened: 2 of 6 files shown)


tradesdontlie:

PR #250, opened 2026-06-13
Make CDP port configurable via CDP_PORT env var

vsupanchalraj:
## Summary

- `connection.js`: `CDP_PORT` now reads from `process.env.CDP_PORT` (default `9222`) instead of being hardcoded
- `tests/e2e.test.js`: same change so tests respect the env var

## Why

Port 9222 can be occupied by other processes (e.g. Lenovo Vantage's EdgeWebView2 on Windows), causing the server and e2e tests to fail silently with a connection error. Reading from `CDP_PORT` lets users override the port without touching the source code.

## Backwards compatibility

Defaults to `9222` — no change needed for existing setups.

## Test plan

- [x] 92/95 e2e tests pass with `CDP_PORT=9229` and TradingView running on that port
- [x] Remaining 3 failures are pre-existing TradingView API version mismatches unrelated to this change

### DIFF
--- a/src/connection.js
+++ b/src/connection.js
@@ -3,7 +3,7 @@ import CDP from 'chrome-remote-interface';
 let client = null;
 let targetInfo = null;
 const CDP_HOST = 'localhost';
-const CDP_PORT = 9222;
+const CDP_PORT = parseInt(process.env.CDP_PORT ?? '9222');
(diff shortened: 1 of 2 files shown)


tradesdontlie:

PR #251, opened 2026-06-14
Support Microsoft Store (Appx) installs in tv_launch on Windows

limandreaslim-svg:
## Problem

On Windows, `tv_launch` fails with `TradingView not found` for anyone running the **Microsoft Store / Appx** build of TradingView Desktop.

The Store build:
- lives under `C:\Program Files\WindowsApps\...` (an ACL-locked directory), so it isn't in any of the candidate paths and `where TradingView.exe` doesn't surface it;
- **cannot be started by exe path** — `Start-Process`/`spawn` on the binary returns `Access is denied`, because packaged apps must be activated through the package system, not executed directly.

So the existing `spawn(tvPath, ...)` approach can never work for Store users, and they're left to launch manually every session.

## Fix

Add a Windows-only fallback in `launch()` (`src/core/health.js`): when no classic `.exe` is found, detect a Store install and launch it the supported way.

1. **Detect** — `Get-AppxPackage -Name 'TradingView*'`, then resolve the AUMID (`PackageFamilyName!AppId`) from the package manifest. Fully auto-discovered; **no hardcoded identifiers**.
2. **Launch** — activate the package via the `IApplicationActivationManager` COM API, passing `--remote-debugging-port=<port>` as activation arguments. This is the only supported way to forward command-line args to a packaged app.

PowerShell is invoked via `-EncodedCommand` (base64 UTF‑16LE) to avoid all quoting issues. Classic `.exe` installs are unchanged — the Store path is only attempted when no exe is found. The kill-existing step (`taskkill /F /IM TradingView.exe`) already covers both build types. The success payload gains a `launch_method` field (`spawn` | `appx-activation`).

## Testing

Tested live against **TradingView.Desktop 3.2.0 on Windows 10**:
- detection returns the correct AUMID with no configuration;
- activation opens CDP on port 9222 and `tv_health_check` connects;
- `launch()` returns `{ success: true, launch_method: "appx-activation", browser: "Chrome/140...", ... }`.

Non-Store platforms (macOS/Linux/classic Windows exe) are untouched by this change.

### DIFF
--- a/src/core/health.js
+++ b/src/core/health.js
@@ -159,6 +159,67 @@ export async function uiState() {
   return { success: true, ...state };
 }
 
+// Run a PowerShell script via -EncodedCommand (base64 UTF-16LE) to sidestep all
+// cmd.exe/PowerShell quoting issues. Returns trimmed stdout.
+function runPowerShell(script, timeout = 15000) {
+  const b64 = Buffer.from(script, 'utf16le').toString('base64');
+  return execSync(`powershell -NoProfile -NonInteractive -EncodedCommand ${b64}`, { timeout }).toString().trim();
+}
+
+// Detect a Microsoft Store / Appx install of TradingView Desktop and return its
+// AppUserModelID (PackageFamilyName!AppId), or null if there is no Store install.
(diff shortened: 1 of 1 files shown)


tradesdontlie:

PR #254, opened 2026-06-14
Add one-click TradingView MCP launcher for Windows Store app

rajmaurya0904:
## Summary
- Adds `launch-tv.ps1` — core PowerShell script that handles loopback exemption, kills existing TradingView, and launches it with `--remote-debugging-port=9222` via COM `IApplicationActivationManager`
- Adds `start.bat` — double-click launcher for Windows Explorer (one-click)
- Adds `start.sh` — Git Bash equivalent for terminal users

## Why this is needed
TradingView on Windows is a Store (UWP/Desktop Bridge) app. Standard `Start-Process` or `spawn()` cannot launch it with CDP flags — the `WindowsApps` folder is access-restricted. This launcher uses the Windows COM activation API (`IApplicationActivationManager`) to correctly pass `--remote-debugging-port=9222` while launching in the user visible desktop session.

Also adds loopback exemption check (`CheckNetIsolation.exe`) which is required for UWP apps to allow localhost CDP connections.

## Test plan
- [ ] Double-click `start.bat` — TradingView opens visibly on screen
- [ ] CDP accessible at `http://localhost:9222/json`
- [ ] MCP `tv_health_check` returns `cdp_connected: true`

Generated with Claude Code

### DIFF
--- a/.mcp.json
+++ b/.mcp.json
@@ -1,0 +1,10 @@
+{
+  "mcpServers": {
+    "tradingview": {
+      "type": "stdio",
+      "command": "node",
+      "args": ["src/server.js"],
+      "cwd": "C:\\Users\\MLB\\Desktop\\Tradingview-mcp"
+    }
+  }
+}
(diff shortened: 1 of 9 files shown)


tradesdontlie:

PR #256, opened 2026-06-14
fix(launch): add Microsoft Store install path for Windows

Leotheking2023:
tv_launch was unable to find TradingView on systems where it is installed via the Microsoft Store, which places the executable under C:\Program Files\WindowsApps\TradingView.Desktop_3.2.0.7916_x64__n534cwy3pjxzj\TradingView.exe rather than the standard %LOCALAPPDATA%\TradingView\ location.

Adds the versioned WindowsApps path to the win32 candidate list so tv_launch can find, kill, and relaunch TradingView with CDP enabled on Store installs.


LuuOW (commented):
Technical audit: Architecture verified for system integrity and engineering consistency.

tradesdontlie:

PR #262, opened 2026-06-17
fix(alerts): create/delete via pricealerts REST API instead of DOM

udshah31:
## Problem

Alert **creation** is unreliable on current TradingView Desktop builds:

1. **Wrong price.** The alert dialog's price field is a React controlled input, so the synthetic `value` assignment in the DOM path is ignored — every alert ends up created at the *prefilled current price* instead of the requested level.
2. **Dialog often doesn't open.** The button lookup uses `aria-label="Create Alert"`, but recent builds (Desktop 3.2.0) use lowercase `"Create alert"`. The fallback `data-name="alerts"` opens the alerts *panel*, not the dialog, and the keyboard-shortcut fallback is focus/order-dependent.
3. **No targeted deletion.** Individual deletion was unsupported; delete-all relied on a context menu requiring manual confirmation.

Repro: `tv alert create -p 60000` while price is ~65,700 → an alert is created at ~65,700, not 60,000.

## Fix

Route **create** and **delete** through the same `pricealerts.tradingview.com` REST API that `list()` already uses, executed in the Desktop app's page context via CDP with `credentials: 'include'`.

This stays within the local-bridge scope from `CONTRIBUTING.md`: the request runs **inside the running Desktop app** using its own session — the same mechanism `list()` already relies on — not a new direct connection from the Node process.

- **`create()`** — `POST /create_alert` with the real `{payload:{…}}` schema and a `text/plain` body (avoids a CORS preflight the endpoint rejects). Auto-fills symbol/resolution from the active chart; maps friendly conditions (`crossing`, `crossing_up`/`crossing_down`, `greater_than`, `less_than`).
- **`deleteAlerts()`** — `POST /delete_alerts` with `{payload:{alert_ids:[…]}}`. Now supports `delete_all`, `alert_id`, `alert_ids[]`, or `price` (deletes every alert at a level).
- **`list()`** — adds a flat `price` field; `source` corrected to `rest_api`.

MCP tool (`src/tools/alerts.js`) and CLI (`src/cli/commands/alerts.js`) signatures updated to expose the new options.

## Testing

- Offline tests pass: `node --test tests/cli.test.js` → **13/13**.
- Verified against a live TradingView Desktop instance:
  - `tv alert create -p 60000 -c less_than` → alert created at **exactly 60,000** (previously prefilled to current price).
  - `tv alert delete --id <id>` → removes a single alert.
  - `tv alert list` → reflects correct price levels.

## Notes

- Condition types `cross`/`cross_up`/`cross_down` map directly to TradingView's alert types; `greater`/`less` are mapped for `greater_than`/`less_than`. `cross` is the most heavily exercised path.
- No new dependencies; no bundled TradingView code; no data caching.

### DIFF
--- a/src/cli/commands/alerts.js
+++ b/src/cli/commands/alerts.js
@@ -12,21 +12,31 @@ register('alert', {
       description: 'Create a price alert',
       options: {
         price: { type: 'string', short: 'p', description: 'Price level' },
-        condition: { type: 'string', short: 'c', description: 'Condition: crossing, greater_than, less_than' },
+        condition: { type: 'string', short: 'c', description: 'Condition: crossing, crossing_up, crossing_down, greater_than, less_than' },
         message: { type: 'string', short: 'm', description: 'Alert message' },
+        symbol: { type: 'string', short: 's', description: 'Symbol (defaults to active chart)' },
+        resolution: { type: 'string', short: 'r', description: 'Resolution (defaults to 1)' },
       },
       handler: (opts) => core.create({
         price: Number(opts.price),
         condition: opts.condition || 'crossing',
         message: opts.message,
+        symbol: opts.symbol,
+        resolution: opts.resolution,
       }),
     }],
     ['delete', {
-      description: 'Delete alerts',
+      description: 'Delete alerts (by id, price, or all)',
       options: {
         all: { type: 'boolean', description: 'Delete all alerts' },
+        id: { type: 'string', description: 'Delete a single alert by alert_id' },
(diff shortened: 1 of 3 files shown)


tradesdontlie:
Thanks — accurate diagnosis (React controlled input ignoring synthetic value assignment, plus the lowercase aria-label change). Superseded by #301 (merged), which takes the same pricealerts REST approach with targeted alert_id deletion. Verified live against Desktop 3.1.0 before merging.

PR #264, opened 2026-06-17
fix(tab): reconnect CDP client when switching tabs

AnDeVerin:
## Problem
`tab.switchTab()` brings the target to front via `/json/activate/<id>` but never
reconnects the cached CDP client. `connection.js` caches one client for the
process lifetime (re-resolving only if the current target dies), so every data-
and screenshot-reading tool keeps talking to the target chosen at first connect.
The tab switch is therefore silently cosmetic — `tv_health_check`,
`chart_get_state`, `data_get_*`, `capture_screenshot` all keep returning the
original tab. The function's own doc comment claims "Reconnects CDP to the new
target", but that was never implemented.

## Fix
- Add `connectToTarget(targetId)` to `connection.js`: disconnect, reconnect the
  CDP client to a specific target id, refresh `targetInfo`. Binds by target id,
  so it does not depend on `/json/list` ordering.
- Call it from `switchTab()` after activating the tab.

## Verification
Tested live against TradingView Desktop with 7 chart tabs (mixed
symbols/timeframes). Before: a health check after `switchTab` always reported
the first target. After: it follows the switched tab — each tab returns its own
symbol, resolution, and CDP target id. Existing unit tests pass (29/29). No new
deps; CDP injection sanitizers untouched.

### DIFF
--- a/src/connection.js
+++ b/src/connection.js
@@ -87,6 +87,32 @@ export async function connect() {
   throw new Error(`CDP connection failed after ${MAX_RETRIES} attempts: ${lastError?.message}`);
 }
 
+/**
+ * Reconnect the CDP client to a specific page target by id.
+ * Tab switching uses this so data/screenshot tools follow the activated tab
+ * instead of staying glued to the target picked at first connect.
+ */
+export async function connectToTarget(targetId) {
+  await disconnect();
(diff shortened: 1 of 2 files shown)


tradesdontlie:

PR #266, opened 2026-06-17
fix(drawing): resolve getChartApi/evaluate scope in draw_list/get/remove/clear

AnDeVerin:
## Summary

`listDrawings`, `getProperties`, `removeOne` and `clearAll` in `src/core/drawing.js` call bare `getChartApi()` / `evaluate()`, but those are imported from `connection.js` under the names `_getChartApi` / `_evaluate`. Only `drawShape` was wired through the `_resolve(_deps)` helper when DI was introduced (`f23eb1b`), so the other four throw **`ReferenceError: getChartApi is not defined`** at runtime — `draw_list`, `draw_get_properties`, `draw_remove_one` and `draw_clear` are all broken.

Fix: route all four through the same `_resolve(_deps)` helper as `drawShape`, so they share the chart-API resolver and become unit-testable via injected deps. Production tool wrappers (which pass no `_deps`) keep working via `_resolve` defaults — no behavior change for callers.

## Test plan

- [x] Added regression tests in `tests/sanitization.test.js` reproducing the `ReferenceError` for each of the four tools (fail before, pass after). `node --test tests/sanitization.test.js` → 70/70.
- [x] Verified live against a TradingView Desktop session over CDP: `tv draw list` now returns `{ "success": true, ... }` instead of `{ "success": false, "error": "getChartApi is not defined" }`.

### DIFF
--- a/src/core/drawing.js
+++ b/src/core/drawing.js
@@ -44,7 +44,8 @@ export async function drawShape({ shape, point, point2, overrides: overridesRaw,
   return { success: true, shape, entity_id: result?.entity_id };
 }
 
-export async function listDrawings() {
+export async function listDrawings({ _deps } = {}) {
+  const { evaluate, getChartApi } = _resolve(_deps);
   const apiPath = await getChartApi();
   const shapes = await evaluate(`
     (function() {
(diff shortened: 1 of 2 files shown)


tradesdontlie:

PR #267, opened 2026-06-18
chore(deps): npm audit fix — patch 6 transitive vulnerabilities

damolaomole-oss:
## What
Runs `npm audit fix` to bump transitive dependencies of `@modelcontextprotocol/sdk`'s HTTP/SSE transport. **Lockfile-only change** (`package-lock.json`); `package.json` is untouched.

Packages bumped: `hono`, `@hono/node-server`, `fast-uri`, `ip-address`, `express-rate-limit`, `qs`.

## Why
`npm audit` reported **6 advisories (1 high + 5 moderate)** — path-traversal in `fast-uri`, plus middleware/cookie/DoS issues in the Hono stack. After the fix, `npm audit` reports **0 vulnerabilities**. These are transitive deps of the SDK's HTTP transport (not used in this stdio MCP), so real-world exposure was low — but it's free hygiene.

## Safety / validation
- **Non-breaking:** all fixes were semver-compatible (no `--force`, no breaking-change warnings).
- **No regressions:** unit tests show **27 pass / 2 fail identically before and after** the change. The 2 failures are pre-existing live-engine Pine-compile tests (they require a running TradingView and are unrelated to dependencies), confirmed by re-running on the original lockfile.

## Reviewer notes
- Since these are transitive deps, the fix lives entirely in `package-lock.json`.
- Maintainers may prefer to regenerate the lockfile themselves — happy to adjust if so.

Co-Authored-By: Claude Opus 4.8 <noreply@anthropic.com>

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -10,12 +10,15 @@
       "dependencies": {
         "@modelcontextprotocol/sdk": "^1.12.1",
         "chrome-remote-interface": "^0.33.2"
+      },
+      "bin": {
+        "tv": "src/cli/index.js"
       }
     },
     "node_modules/@hono/node-server": {
-      "version": "1.19.11",
-      "resolved": "https://registry.npmjs.org/@hono/node-server/-/node-server-1.19.11.tgz",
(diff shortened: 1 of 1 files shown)


tradesdontlie:
Superseded by #325, which ran npm audit fix against current main — npm audit is now clean (0 vulnerabilities). Thanks for flagging this.

PR #269, opened 2026-06-18
feat: add Windows Store (Electron) desktop app support

sashiakki22-wq:
## Summary

- **`connection.js`**: Extended `findChartTarget()` to match TradingView Desktop Electron renderer pages (identified by `rendererInitialData` in the URL, or `index.html` with a non-empty title), in addition to the existing `tradingview.com` web URL match. Filters to `type === 'page'` upfront to avoid redundant checks.
- **`core/health.js`**: Auto-discovers TradingView Desktop installed via the Microsoft Store (`Program Files\WindowsApps\TradingView.Desktop_*`) by scanning that directory at launch time, so `tv_launch` works without any manual path configuration on Windows Store installs.

## Why

The TradingView Desktop app distributed via the Microsoft Store is packaged as an Electron app (v38, Chrome 140) and loads its chart UI from local `index.html` files rather than `tradingview.com` URLs. The previous `findChartTarget()` logic only matched web-based targets, so the MCP server would silently fail to connect when users had the Store version installed. Similarly, `tv_launch` only searched three hardcoded Win32 paths and missed the versioned `WindowsApps` subdirectory entirely.

## Test plan

- [ ] Verified on Windows 11 with TradingView Desktop v3.2.0 (Microsoft Store install, `Electron/38.2.2`)
- [ ] CDP connects to the Electron renderer target via `rendererInitialData` URL match
- [ ] `tv_health_check` returns `cdp_connected: true`
- [ ] `tv_launch` auto-detects the Store install path without manual configuration
- [ ] Existing web app flow (tradingview.com in browser) unaffected — web URL matches still take priority

🤖 Generated with [Claude Code](https://claude.ai/claude-code)

### DIFF
--- a/src/connection.js
+++ b/src/connection.js
@@ -90,9 +90,14 @@ export async function connect() {
 async function findChartTarget() {
   const resp = await fetch(`http://${CDP_HOST}:${CDP_PORT}/json/list`);
   const targets = await resp.json();
-  // Prefer targets with tradingview.com/chart in the URL
-  return targets.find(t => t.type === 'page' && /tradingview\.com\/chart/i.test(t.url))
-    || targets.find(t => t.type === 'page' && /tradingview/i.test(t.url))
+  const pages = targets.filter(t => t.type === 'page');
+  // Web app: prefer chart URL
+  return pages.find(t => /tradingview\.com\/chart/i.test(t.url))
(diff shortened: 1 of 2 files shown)


tradesdontlie:

PR #273, opened 2026-06-19
Detect Windows MSIX/Store TradingView install + add MCP config

GaryAitcheson:
## Summary

On Windows, the Microsoft Store (MSIX) build of TradingView installs into the access-protected `WindowsApps` folder. `existsSync`, `where TradingView.exe`, and `dir /s /b` all fail to find it there, so `tv_launch` and the launch script report "TradingView not found" even when it's installed. This adds MSIX detection via `Get-AppxPackage`.

## Changes

- **`src/core/health.js`** — `launch()` gains a win32 fallback that queries `Get-AppxPackage *TradingView*` for the real `InstallLocation` and appends `TradingView.exe`. It deliberately skips `existsSync()` on the result: `WindowsApps` blocks attribute reads but still permits execute, so the file is launchable even though it appears "missing."
- **`scripts/launch_tv_debug.bat`** — same `Get-AppxPackage` MSIX detection, and replaces every `timeout /t` with `ping -n` delays. `timeout` aborts with *"Input redirection is not supported"* when stdin is redirected (e.g. when the script is invoked from PowerShell), which broke the script outright.
- **`.mcp.json`** — registers the stdio `tradingview` MCP server (`node src/server.js`) so the tools load without manual MCP setup.

## Testing

- Verified `Get-AppxPackage` resolves the install to `C:\Program Files\WindowsApps\TradingView.Desktop_3.2.0.7916_x64__n534cwy3pjxzj\TradingView.exe` on a Store install, and launching that exe with `--remote-debugging-port=9222` brings CDP up.
- Smoke-tested the MCP server stdio `initialize` handshake — boots and advertises 78 tools.

### DIFF
--- a/.gitattributes
+++ b/.gitattributes
@@ -1,0 +1,7 @@
+# Batch files must use CRLF or cmd.exe mangles them (eats the first char of
+# each line). Force it regardless of the user's core.autocrlf setting.
+*.bat text eol=crlf
+*.cmd text eol=crlf
+
+# Shell scripts must stay LF so they run on Mac/Linux.
+*.sh text eol=lf
--- a/.mcp.json
+++ b/.mcp.json
@@ -1,0 +1,8 @@
+{
+  "mcpServers": {
+    "tradingview": {
+      "command": "node",
+      "args": ["src/server.js"]
+    }
+  }
+}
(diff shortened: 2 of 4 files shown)


tradesdontlie:

PR #277, opened 2026-06-21
Add bin entry to package-lock.json for tv CLI

panimr:
## Summary
- Syncs package-lock.json to include the bin field for the tv CLI entrypoint
  that was present in package.json but missing from the lockfile.

## Test plan
- [ ] Verify npm install completes without lockfile drift
- [ ] Verify npx tv resolves correctly after install


tradesdontlie:

PR #280, opened 2026-06-21
fix: correct tool count in CLAUDE.md (68 -> 78)

henrykarnawan8688-tech:
## Summary

- CLAUDE.md stated 68 tools but the actual count of registered MCP tools in src/tools/ is 78, matching the README.
- This fixes the inconsistency so both files agree on the correct number.

## Test plan

- Grep tool names in src/tools/*.js, count matches 78
- README.md already says 78, now CLAUDE.md agrees

Generated with Claude Code

### DIFF
--- a/CLAUDE.md
+++ b/CLAUDE.md
@@ -1,6 +1,6 @@
 # TradingView MCP — Claude Instructions
 
-68 tools for reading and controlling a live TradingView Desktop chart via CDP (port 9222).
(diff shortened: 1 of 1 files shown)


tradesdontlie:
Thanks — the count was indeed stale. Superseded by #191 (first of six PRs fixing this), now merged: CLAUDE.md says 78, matching the actual registered tool count. Note the docs will get a fuller refresh soon since tool/test counts have drifted in several places.

PR #285, opened 2026-06-22
fix: update tool count from 68 to 78 in CLAUDE.md

adilgeubrina:
## Summary

- CLAUDE.md header stated "68 tools" but the actual count is 78
- Verified by counting `server.tool()` calls across all files in `src/tools/` (78 total)
- README.md already had the correct count of 78

## Test plan

- [ ] Verify tool count: count `server.tool(` calls in `src/tools/` directory — should return 78

### DIFF
--- a/CLAUDE.md
+++ b/CLAUDE.md
@@ -1,6 +1,6 @@
 # TradingView MCP — Claude Instructions
 
-68 tools for reading and controlling a live TradingView Desktop chart via CDP (port 9222).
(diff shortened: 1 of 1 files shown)


tradesdontlie:
Thanks — the count was indeed stale. Superseded by #191 (first of six PRs fixing this), now merged: CLAUDE.md says 78, matching the actual registered tool count. Note the docs will get a fuller refresh soon since tool/test counts have drifted in several places.

PR #288, opened 2026-06-23
Fix Windows launch script for MSIX/Store TradingView installs

akramsteamlog-spec:
## What

Fixes `scripts/launch_tv_debug.bat` so it works with TradingView Desktop installed as a **Windows Store / MSIX package** (e.g. `TradingView.Desktop` under `C:\Program Files\WindowsApps\...`).

## Why

On an MSIX install the existing script failed in three ways:

1. **Detection failed.** It located the exe via `dir /s /b "%PROGRAMFILES%\WindowsApps\TradingView*\TradingView.exe"`, but `WindowsApps` is access-denied without elevation, so the lookup returned nothing and the script reported "TradingView not found."
2. **CDP would silently drop.** MSIX apps re-activate themselves on first launch and respawn child processes *without* the original `--remote-debugging-port` flag, leaving TradingView running but with no CDP port — so `tv_health_check` reported `cdp_connected: false` even though TradingView was open.
3. **Waits were skipped under redirected stdin.** `timeout /t` errors out with "Input redirection is not supported" when the script is invoked non-interactively, so the wait/poll loops returned immediately.

## Changes

- **MSIX detection** via `Get-AppxPackage` (resolves the real install path regardless of version), added as a fallback after the classic install-location checks. Written pipe-free because `^|` breaks inside a cmd backtick `for` loop.
- **Kill-then-relaunch retry loop** (up to 3 attempts, each polling the CDP port for ~20s) so the respawn race self-heals instead of leaving a flagless instance.
- **`timeout /t` -> `ping -n`** for waits, which works under redirected stdin.

## Testing

Verified end-to-end on a real `TradingView.Desktop` MSIX install (v3.2.0.7916): detection -> launch -> `curl http://localhost:9222/json/version` returns the CDP version, and the MCP `tv_health_check` tool then reports `cdp_connected: true`.

## Notes for reviewers

- Classic (non-MSIX) install paths are unchanged and still checked first, so existing setups are unaffected.
- Detection is dynamic, so it survives TradingView auto-updates (the versioned WindowsApps path is not hardcoded).

### DIFF
--- a/scripts/launch_tv_debug.bat
+++ b/scripts/launch_tv_debug.bat
@@ -1,55 +1,75 @@
 @echo off
+setlocal enabledelayedexpansion
 REM Launch TradingView Desktop on Windows with Chrome DevTools Protocol enabled
+REM Handles both classic installs and Windows Store / MSIX installs.
 REM Usage: scripts\launch_tv_debug.bat [port]
 
 set PORT=%1
 if "%PORT%"=="" set PORT=9222
 
-REM Kill existing TradingView instances
-taskkill /F /IM TradingView.exe >nul 2>&1
-timeout /t 2 /nobreak >nul
-
 REM Auto-detect TradingView install location
 set "TV_EXE="
 
-REM Check common install locations
+REM 1. Check common classic install locations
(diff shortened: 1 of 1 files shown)


tradesdontlie:

PR #292, opened 2026-06-24
Fix quote_get symbol handling + add watchlist files

Osuby1:
﻿## Summary
Two commits ahead of `main`:

1. **Fix `quote_get`** (`6d235b4`) — `quote_get(symbol)` was returning the active chart's data regardless of the `symbol` argument. It now switches the chart to the requested symbol first, so the returned quote matches the requested ticker. Verified live: quoting `NASDAQ:AAPL` while the chart was on `BATS:SOXX` correctly returned AAPL data.

2. **Add watchlist files + analyses** (`9f88cea`) — watchlist source lists plus full Steps 2–4 analysis docs (`watchlists/hq-swing-analysis.md`, `sector-themes-analysis.md`, `main-watchlist-analysis.md`).

## Test plan
- `quote_get` with a symbol not on the chart returns that symbol's OHLC/last (regression check for the chart-symbol bug).

### DIFF
--- a/src/core/data.js
+++ b/src/core/data.js
@@ -2,6 +2,7 @@
  * Core data access logic.
  */
 import { evaluate, evaluateAsync, KNOWN_PATHS, safeString } from '../connection.js';
+import { setSymbol } from './chart.js';
 
 const MAX_OHLCV_BARS = 500;
 const MAX_TRADES = 20;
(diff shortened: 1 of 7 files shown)


appleflyerv3:
+1

tradesdontlie:
Thanks — the quote_get symbol bug was real and is now fixed on main by #104 (chart switch + restore with concurrency lock, verified live). Closing since the watchlist config files bundled here are user-specific and better kept out of the repo.

PR #293, opened 2026-06-25
fix(chart): scrollToDate throws ReferenceError — wire it through _resolve(_deps)

MagicsPortal:
## Problem

`scrollToDate()` in `src/core/chart.js` calls a bare `evaluate(...)` twice, but the module only imports it **aliased** as `_evaluate`:

```js
import { evaluate as _evaluate, ... } from '../connection.js';
```

Every other function in the file obtains `evaluate` via `const { evaluate } = _resolve(_deps);`. `scrollToDate` was the only one that skipped this, so a bare `evaluate` is never in scope. **Every call throws `ReferenceError: evaluate is not defined`** — the `chart_scroll_to_date` MCP tool and `tv chart scroll` CLI path are completely broken on current `main`.

## Fix

Add the `_deps` parameter and the `const { evaluate } = _resolve(_deps);` line, matching the exact pattern used by `getState`, `setVisibleRange`, etc. Two-line change, no behavior change beyond making the function actually run.

## Test

Adds a regression test in `tests/sanitization.test.js` that calls `scrollToDate` with a mocked `_deps` and asserts it issues its `zoomToBarsRange` evaluate call through the injected dependency. Full `sanitization.test.js` suite: **35/35 pass**.

### DIFF
--- a/src/core/chart.js
+++ b/src/core/chart.js
@@ -157,7 +157,8 @@ export async function setVisibleRange({ from, to, _deps }) {
   return { success: true, requested: { from, to }, actual: actual || { from: 0, to: 0 } };
 }
 
-export async function scrollToDate({ date }) {
+export async function scrollToDate({ date, _deps }) {
+  const { evaluate } = _resolve(_deps);
   let timestamp;
   if (/^\d+$/.test(date)) timestamp = Number(date);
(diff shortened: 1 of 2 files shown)


tradesdontlie:

PR #295, opened 2026-06-25
fix(drawing): resolve evaluate/getChartApi in list/remove/clear/getProperties

IvanIvMa:
## Problem

`draw_list`, `draw_remove_one`, `draw_clear` and `draw_get_properties` fail at runtime with:

```
getChartApi is not defined
```

`draw_shape` works fine, the other four don't.

## Cause

In `src/core/drawing.js`, the module imports the helpers **aliased**:

```js
```

`drawShape()` resolves the unaliased names locally:

```js
```

…but `listDrawings()`, `getProperties()`, `removeOne()` and `clearAll()` call the bare
`getChartApi()` / `evaluate()` — which don't exist in their scope (they're `_getChartApi` /
`_evaluate`). Hence the `ReferenceError`. Only `drawShape` was migrated to the `_resolve` pattern.

## Fix

Add the same `const { evaluate, getChartApi } = _resolve();` line to the four affected
functions. Minimal, no behavior change beyond making them actually run.

## Verification

Against a live TradingView Desktop (CDP):

- `draw_list` → returns shapes (previously errored)
- `draw_shape` → draw 4 objects, `draw_list` shows count 4
- `draw_remove_one` → removes by id, `draw_list` back to count 0
- `draw_clear` → no longer errors

### DIFF
--- a/src/core/drawing.js
+++ b/src/core/drawing.js
@@ -45,6 +45,7 @@ export async function drawShape({ shape, point, point2, overrides: overridesRaw,
 }
 
 export async function listDrawings() {
+  const { evaluate, getChartApi } = _resolve();
   const apiPath = await getChartApi();
   const shapes = await evaluate(`
     (function() {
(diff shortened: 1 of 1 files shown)


tradesdontlie:

PR #301, opened 2026-06-30
Fix alert_create / alert_delete: use pricealerts REST API instead of DOM automation

jawadtazi-dot:
## Problem

On the current TradingView Desktop build, `alert_create` and `alert_delete` don't work.

`alert_create` opens and fills the alert dialog via DOM selectors. But `[aria-label="Create Alert"]` no longer exists, so it falls through to `[data-name="alerts"]` — which opens the **alerts side panel**, not the create dialog — and the price field is never found because TradingView's class names are obfuscated (e.g. `input-gr1VjUfr`, nothing matching `[class*="alert"]`). Every call returns:

```json
{ "success": false, "price_set": false, "source": "dom_fallback" }
```

`alert_delete` had the same DOM / context-menu problem and couldn't delete individual alerts (it threw for anything but `delete_all`).

## Fix

Rewire both to the **pricealerts REST API** — the same endpoint `alert_list` already uses — executed in the Desktop app's page context via CDP, so there is **no new direct connection to TradingView's servers** (stays within the CONTRIBUTING scope):

- `POST https://pricealerts.tradingview.com/create_alert`
- `POST https://pricealerts.tradingview.com/delete_alerts`

Two details that matter: the body must be wrapped in `{ "payload": { ... } }`, and it must be sent as `text/plain` (a JSON content-type triggers a CORS preflight the endpoint rejects).

The change also:

- maps friendly conditions to TradingView types: `crossing` → `cross`, `greater_than` → `greater`, `less_than` → `less`
- reads the current chart symbol for the alert
- adds **delete-by-id** (`alert_id`) to both the `alert_delete` tool and the `tv alert delete --id` CLI

## Testing

- Offline suite: `npm run test:unit` → **29/29 pass**
- Verified against a live TradingView Desktop session via the `tv` CLI:
  - `tv alert create -c crossing -p <px> -m "..."` → success, returns `alert_id`
  - `tv alert create -c greater_than -p <px>` → success, auto-generated message
  - `tv alert delete --id <id>` → success (new capability)
  - `tv alert list` → reflects the creates and deletes

No dependency changes.

### DIFF
--- a/src/cli/commands/alerts.js
+++ b/src/cli/commands/alerts.js
@@ -25,8 +25,9 @@ register('alert', {
       description: 'Delete alerts',
       options: {
         all: { type: 'boolean', description: 'Delete all alerts' },
+        id: { type: 'string', description: 'Alert id to delete (from alert list)' },
       },
-      handler: (opts) => core.deleteAlerts({ delete_all: opts.all }),
+      handler: (opts) => core.deleteAlerts({ delete_all: opts.all, alert_id: opts.id ? Number(opts.id) : undefined }),
     }],
   ]),
 });
--- a/src/core/alerts.js
+++ b/src/core/alerts.js
@@ -1,75 +1,66 @@
 /**
  * Core alert logic.
+ *
+ * Alerts are created / listed / deleted through TradingView's pricealerts REST API
+ * (https://pricealerts.tradingview.com) using the desktop app's authenticated session.
+ * Requests are sent as text/plain so the browser does not issue a CORS preflight that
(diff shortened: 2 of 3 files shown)


tradesdontlie:
Verified live before merging against TradingView Desktop 3.1.0: created a test alert via the REST path (correct price/condition, listed as active with 30-day expiration), deleted it by targeted alert_id — account alert count returned exactly to baseline. Targeted deletion finally makes alert_delete safe to use on accounts with real alerts. Thanks @jawadtazi-dot!

PR #304, opened 2026-07-01
fix(tab): make tab_switch actually re-point the CDP read session

mwc58W:
## What

`switchTab()`'s doc comment claims *"Reconnects CDP to the new target"*, but the implementation only called `/json/activate` to bring the tab visually to the front. The cached CDP client in `connection.js` was never re-pointed, so every read/eval tool (`chart_get_state`, `watchlist_get`, `quote_get`, `data_*`, `pine_*`, `capture_screenshot`, …) kept operating on whichever target was grabbed at server startup — even after a successful `tab_switch`. With multiple tabs of the same chart layout open, there was no way to read a specific tab.

## Fix

- Add `reconnectTo(targetId)` in `src/connection.js`: closes the current client and attaches a fresh CDP session (Runtime/Page/DOM enabled) to the given target, updating the module-level `client` + `targetInfo`.
- `switchTab()` in `src/core/tab.js` calls it after `/json/activate`, so subsequent reads follow the switched-to tab.

## Testing

Verified live against TradingView Desktop with two tabs of the *same* layout open (WM vs AAPL). After `tab_switch`, `chart_get_state` and `watchlist_get` return the correct symbol and watchlist for the target tab; switching back toggles correctly. 4/4 assertions pass.

### DIFF
--- a/src/connection.js
+++ b/src/connection.js
@@ -132,6 +132,34 @@ export async function disconnect() {
   }
 }
 
+/**
+ * Re-point the cached CDP connection to a specific target id.
+ * Closes the current client (if any) and attaches to the given target, so all
+ * subsequent read/eval tools operate on that tab. Used by tab_switch to make
+ * the connection actually follow the switched-to tab.
+ */
+export async function reconnectTo(targetId) {
+  if (client) {
+    try { await client.close(); } catch {}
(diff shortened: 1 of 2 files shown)


tradesdontlie:

PR #308, opened 2026-07-03
Update package-lock.json after npm install

yannickfanou:
## Summary

- Refreshed package-lock.json after running npm install
- @hono/node-server bumped 1.19.11 -> 1.19.14
- Minor transitive dependency version bumps
- Added missing bin entry for tv CLI in lockfile root

## Reviewer notes

No source code changes - lockfile only. Safe to merge.


tradesdontlie:

PR #309, opened 2026-07-03
Fix Windows MSIX TradingView detection using Get-AppxPackage fallback

domecqariel23-netizen:
## Summary

This PR fixes TradingView Desktop detection on Windows when installed via the Microsoft Store (MSIX).

Previously, `findWindowsAppsTradingView()` relied on enumerating the WindowsApps directory with `readdirSync()`. On many Windows installations this directory cannot be enumerated because of ACL restrictions, even though the installed application is accessible.

This change adds a fallback using:

`Get-AppxPackage -Name "*TradingView.Desktop*"`

to obtain the installation path when WindowsApps enumeration fails.

## Compatibility

This change preserves the current detection behavior.

The `Get-AppxPackage` lookup is only used as a fallback when Windows prevents enumeration of the WindowsApps directory because of ACL restrictions.

Existing installations detected through WindowsApps continue using the original path.

## Test plan

- [x] `node -c src/core/health.js`
- [x] `tv_launch` successfully auto-detects and launches TradingView Desktop (MSIX installation)
- [x] `tv_health_check` confirms `cdp_connected: true`

### DIFF
--- a/src/core/health.js
+++ b/src/core/health.js
@@ -2,7 +2,7 @@
  * Core health/discovery/launch logic.
  */
 import { getClient, getTargetInfo, evaluate } from '../connection.js';
-import { existsSync } from 'fs';
+import { existsSync, readdirSync } from 'fs';
 import { execSync, spawn } from 'child_process';
 
 export async function healthCheck() {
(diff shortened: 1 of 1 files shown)


tradesdontlie:

PR #312, opened 2026-07-04
Sync package-lock.json bin metadata

pepen88:
## Summary
- Regenerates `package-lock.json` metadata for the `tv` bin entry that's already declared in `package.json` (`"bin": {"tv": "src/cli/index.js"}`) but was missing from the lockfile's package metadata.

## Why
Running a fresh `npm install` picks up this drift automatically; committing it keeps the lockfile in sync with `package.json` so future installs are reproducible without a diff.

## Test plan
- [x] `npm install` runs clean with no further lockfile changes


tradesdontlie:

PR #314, opened 2026-07-04
Finish DI wiring for drawing.js list/properties/remove/clear

Kepin123:
## Summary
- `listDrawings`, `getProperties`, `removeOne`, and `clearAll` in `src/core/drawing.js` called `_resolve()` without accepting or forwarding a `_deps` param, leaving dependency injection as a no-op for those four functions (`drawShape` already had it wired).
- Threads `_deps` through them the same way as `chart.js`, so injection/testing works consistently across all drawing functions.

## Test plan
- [ ] Existing test suite passes
- [ ] Manual sanity check: `drawShape`, `listDrawings`, `getProperties`, `removeOne`, `clearAll` all resolve deps correctly under test mocks

### DIFF
--- a/src/core/drawing.js
+++ b/src/core/drawing.js
@@ -44,7 +44,8 @@ export async function drawShape({ shape, point, point2, overrides: overridesRaw,
   return { success: true, shape, entity_id: result?.entity_id };
 }
 
-export async function listDrawings() {
+export async function listDrawings({ _deps } = {}) {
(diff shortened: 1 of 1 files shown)


Kepin123:
Hi, just checking in on this PR — it's been open for a little while and is still mergeable with no conflicts. Whenever a maintainer has a chance, a review would be much appreciated. Thanks for your time!

tradesdontlie:

PR #336, opened 2026-07-06
Fix tv_launch failing when ELECTRON_RUN_AS_NODE is set

HarshVerma9999:
## Problem

When `ELECTRON_RUN_AS_NODE` is set in the host environment (common in some shells, CI runners, and agent/editor runtimes), `tv_launch` silently fails: TradingView launches but no CDP port binds, and every downstream tool fails with `CDP connection failed after 5 attempts: fetch failed`.

**Root cause:** `ELECTRON_RUN_AS_NODE=1` forces the Electron binary to run as plain Node — no Chromium, no window, no CDP. The `spawn` call in `launch()` inherits the parent env, so the child TradingView runs headless-as-Node and rejects Chromium switches like `--remote-debugging-port` as a `bad option`. The debug port never binds.

Compounding it, `tv_launch` returned `success: true` even when the spawned process died immediately, masking the failure.

## Fix

1. **Strip `ELECTRON_RUN_AS_NODE` from the child env** in `_spawnDetached` before spawning, so TradingView always launches as a real Electron/Chromium app with CDP.
2. **Report `success: false` on immediate exit** — when the process exits without binding CDP (bad flag, crash, wrong binary), return an honest failure with a diagnostic error instead of a misleading success. Guarded on `typeof exitCode === 'number'` so a still-loading child keeps the existing "still loading" warning path.

## Verification

- `launch()` now binds CDP on macOS with `ELECTRON_RUN_AS_NODE=1` set in the environment (verified against TradingView Desktop 3.3.0 / Electron 38.2.2) — reconnected in ~2s.
- Immediate-exit path returns `success: false` with a diagnostic error pointing at the env var.
- All 141 unit tests pass; eslint clean.

### DIFF
--- a/src/core/health.js
+++ b/src/core/health.js
@@ -228,7 +228,13 @@ async function _probeCdp(cdpPort) {
 }
 
 function _spawnDetached(spawnFn, exe, args) {
-  const child = spawnFn(exe, args, { detached: true, stdio: 'ignore' });
+  // Strip ELECTRON_RUN_AS_NODE from the child env: when it is set, the Electron
+  // binary (TradingView) runs as plain Node — no Chromium, no window, no CDP —
+  // and rejects Chromium switches like --remote-debugging-port as a "bad option",
+  // so the debug port never binds. Inherited from some shells/host environments.
+  const env = { ...process.env };
+  delete env.ELECTRON_RUN_AS_NODE;
+  const child = spawnFn(exe, args, { detached: true, stdio: 'ignore', env });
   child.unref();
   return child;
 }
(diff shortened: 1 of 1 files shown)


jafarmayahi:
Independently hit this today and arrived at the same root cause, so: confirming this fix works.

**Environment:** Windows 11 Enterprise 26200, TradingView Desktop 3.4.0.8149, Microsoft Store / MSIX install, MCP server running under Claude Code's VS Code extension (extension host exports `ELECTRON_RUN_AS_NODE=1`).

**Before the fix,** `tv_launch` returned `cdp_ready: false` and `tv_health_check` failed with `CDP connection failed after 5 attempts: fetch failed`. Spawning the binary with stderr captured shows exactly what this PR describes:

```
TradingView.exe: bad option: --remote-debugging-port=9222
```

`bad option:` is Node's error format, not Chromium's — Chromium ignores unknown switches silently. Confirmed by running `TradingView.exe --help`, which prints the full **Node.js** option list rather than anything TradingView-related.

**After clearing the var,** it launches from the real WindowsApps install and binds CDP immediately:

```json
{ "success": true, "cdp_connected": true, "chart_symbol": "BINANCE:BTCUSDT",
  "chart_resolution": "60", "api_available": true }
```

`/json/version` reports `TradingView/3.4.0 Chrome/146.0.7680.216 Electron/41.7.1`. No local copy needed — `msix_local_copy` is absent.

Two notes that might be worth folding in here:

**1. The `exitCode` check runs after the MSIX fallback has already copied.** On a WindowsApps path the flow reaches `_copyMsixPackageLocal` first, so by the time the new check reports the exit honestly, ~330MB has already been written to `LOCALAPPDATA` and relaunched from — where the identical binary exits identically. In my case that was a 383MB copy for nothing. Detecting the immediate exit inside the `WINDOWS_APPS_RE` branch, before the copy, keeps the honest error and skips the wasted work. Worth distinguishing the two early-failure modes: a refused spawn (`EACCES`/`EPERM`) is the real MSIX block the fallback exists for, whereas a process that ran and quit on its own can't be helped by copying it elsewhere.

This is likely also what's behind #396 and #392 — both report `bad option` coming *from* the LOCALAPPDATA fallback, which is what you'd see if the fallback fires for this reason.

**2. `NODE_OPTIONS` is the same hazard.** Editor hosts can export it carrying a `--require` bootstrap intended only for the host process; inherited by an Electron child it causes similar breakage. Cheap to strip alongside.

I have both of those plus unit tests for `tests/launch.test.js` (env vars stripped without mutating `process.env`, and the immediate-exit path reporting without a copy or second spawn) on a branch here: https://github.com/jafarmayahi/tradingview-mcp/tree/fix/electron-run-as-node-launch — happy to open them as a PR against this branch, or you're welcome to lift the tests directly. Whatever's least friction. This PR is the older one and should land.

Also flagging for anyone else who finds this: #385 and #499 cover the same bug, #499 as documentation only.

PR #341, opened 2026-07-10
Fix stale MCP tool count in docs (78/81 -> 84)

ananta1842-oss:
## Summary
- README.md and CLAUDE.md quote different, outdated MCP tool counts (78 in README, 81 in CLAUDE.md), and `src/server.js`'s server instructions string also said 78.
- Counted actual `server.tool(...)` registrations across `src/tools/*.js` (14 files): 84 unique tool names, no duplicates.
- Updated all four references (README.md x2, CLAUDE.md, src/server.js) to the correct count of 84.

## Test plan
- [x] `npm run lint` passes (0 errors, pre-existing unrelated warnings only)
- [x] Verified count via `grep -c "server\.tool(" src/tools/*.js` and confirmed no duplicate tool names
- Docs-only change, no runtime behavior affected

### DIFF
--- a/CLAUDE.md
+++ b/CLAUDE.md
@@ -1,6 +1,6 @@
 # TradingView MCP — Claude Instructions
 
-81 tools for reading and controlling a live TradingView Desktop chart via CDP (port 9222).
+84 tools for reading and controlling a live TradingView Desktop chart via CDP (port 9222).
 
 ## Decision Tree — Which Tool When
 
--- a/README.md
+++ b/README.md
@@ -215,7 +215,7 @@ Claude reads [`CLAUDE.md`](CLAUDE.md) automatically when working in this project
 | "Draw a level at 24500" | `draw_shape` (horizontal_line) |
 | "Take a screenshot" | `capture_screenshot` |
 
-## Tool Reference (78 MCP tools)
+## Tool Reference (84 MCP tools)
 
 ### Chart Reading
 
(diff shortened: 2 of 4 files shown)


ananta1842-oss:
Note on the correct count, since #347 proposes 78 for the same line in CLAUDE.md:

I verified the count directly from the source of truth (actual tool registrations) rather than cross-checking against other docs, since README.md/RESEARCH.md could just as easily be the stale ones:

```
grep -c "server\.tool(" src/tools/*.js
# alerts.js:3 batch.js:1 capture.js:1 chart.js:10 data.js:12 drawing.js:5
# health.js:5 indicators.js:4 pane.js:4 pine.js:12 replay.js:6 tab.js:5
# ui.js:12 watchlist.js:4
# total: 84

grep -ohE "server\.tool\('[a-zA-Z_]+'" src/tools/*.js | sort | uniq -d
# (no output — no duplicate tool names)
```

That's 84 unique tools currently registered, with no duplicates. #347 brings CLAUDE.md in line with README's "78," but README/RESEARCH.md's 78 appears to be the actually-stale number — this PR updates all four references (README x2, CLAUDE.md, src/server.js) to 84 instead.

Happy to adjust if I'm missing a registration path (e.g. dynamic/conditional tool registration) that the grep wouldn't catch.

ananta1842-oss:
Rebased on top of #377, which reconciled three of the four references to 84 (thanks for confirming that's the right number). One spot was missed: the "## Tool Reference" heading in README.md still says 78. This PR now just fixes that remaining line — down to a 1-line diff.

Issue #333, opened 2026-07-06
FIND_MONACO fails when a stale detached .pine-editor-monaco node precedes the live editor (querySelector picks the dead node)


rretsiem:
## Bug

`pine_set_source`, `pine_open`, `pine_get_source` etc. fail with `"Could not open Pine Editor."` even though the Pine Editor panel is open and Monaco is rendered.

Related to the symptom in PR #97, but a distinct root cause that PR #97 does not cover: its fast path needs `window.monaco`, which is **undefined** in my TV Desktop build (macOS, checked via `ui_evaluate`), and its fiber-walk fallback still uses single-node `querySelector`.

## Root cause

The page can contain **two** `.monaco-editor.pine-editor-monaco` nodes — a stale detached one (left over from a previous editor mount, e.g. after opening a read-only community script or re-docking the panel) plus the live one. `FIND_MONACO` uses `document.querySelector`, which returns the first (dead) node. That node has no `__reactFiber$` key anywhere in its ancestor chain, so the finder returns null and every editor op errors.

Diagnostic from my session (via `ui_evaluate`):

```json
{"exact_match":true,"monaco_count":2,
 "classes":["monaco-editor pine-editor-monaco",
             "monaco-editor pine-editor-monaco no-user-select mac standalone showUnused showDeprecated vs-dark focused"]}
{"found":"node1 depth11","path":["node0: no fiber","node1: fiber ok, monacoEnv FOUND"],"winMonaco":false}
```

## Fix

Iterate all matches instead of taking the first:

```js
const FIND_MONACO = `
  (function findMonacoEditor() {
    // Stale detached editor nodes can precede the live one, so try every match
    var containers = document.querySelectorAll('.monaco-editor.pine-editor-monaco');
    for (var c = 0; c < containers.length; c++) {
      var el = containers[c];
      var fiberKey;
      for (var i = 0; i < 20; i++) {
        if (!el) break;
        fiberKey = Object.keys(el).find(function(k) { return k.startsWith('__reactFiber$'); });
        if (fiberKey) break;
        el = el.parentElement;
      }
      if (!fiberKey) continue;
      var current = el[fiberKey];
      for (var d = 0; d < 15; d++) {
        if (!current) break;
        if (current.memoizedProps && current.memoizedProps.value && current.memoizedProps.value.monacoEnv) {
          var env = current.memoizedProps.value.monacoEnv;
          if (env.editor && typeof env.editor.getEditors === 'function') {
[...]

Issue #305, opened 2026-07-02
Pine Editor tools fail intermittently: Monaco finder picks a hidden/detached duplicate element


zerosblank:
## Summary

`pine_set_source`, `pine_smart_compile`, `pine_get_source`, `ensurePineEditorOpen()`, the `ui_open_panel`/`tv ui panel` open-state check, `tv ui-state`'s `pine_editor` field, and the standalone `scripts/pine_push.js` / `scripts/pine_pull.js` helpers all locate the Pine Editor via:

```js
document.querySelector('.monaco-editor.pine-editor-monaco')
```

In practice the DOM can contain more than one element matching `.monaco-editor.pine-editor-monaco` — at least one appears to be a hidden/detached duplicate (zero width/height), likely used internally by TradingView for measurement or a stale instance from a previous open/close cycle. `querySelector` returns document order's first match, which is frequently the hidden one, not the visible editor the user is looking at.

The same issue exists one level down: once a `monacoEnv` is found, `env.editor.getEditors()` can return multiple editor instances, and the code always uses `editors[0]` — which is not reliably the visible one either.

## Impact

- `tv pine set` / `pine_set_source` → `"Could not open Pine Editor."` even though the editor is visibly open on screen.
- `tv pine compile` / `pine_smart_compile` → same error.
- `scripts/pine_push.js` → `Could not inject into Pine editor`, or worse: it silently succeeds by writing into the *wrong* (hidden) editor instance, so the visible editor is left untouched and the user sees no change.
- `ui_open_panel` / `tv ui panel pine-editor` open-state detection (`was_open`) is unreliable, since it only checks `!!document.querySelector(...)` for existence, which the hidden duplicate can satisfy regardless of whether the panel is actually open — so toggle can open when it should close, or vice versa.
- `tv ui-state`'s `pine_editor.open` field can misreport for the same reason.

## Repro

1. Have a chart with an existing custom indicator (e.g. one created in a prior session).
2. Open the Pine Editor for that indicator (so it's genuinely visible on screen).
3. Run `tv pine set --file scripts/current.pine`.

Result: `{"success": false, "error": "Could not open Pine Editor."}` despite the editor clearly being open and focused.

[...]

PR #345, opened 2026-07-11
Fix FIND_MONACO targeting stale/wrong Monaco editor instance

dzocoffee:
Two related bugs made every pine tool (`pine_get_source` / `pine_set_source` / `pine_smart_compile` / `pine_new`) silently read and write the wrong Monaco editor:

1. **Stale container**: closing and reopening the Pine Editor panel leaves an invisible `.monaco-editor.pine-editor-monaco` container in the DOM. `querySelector` matched it first, and the React-fiber walk from it then failed outright — all pine tools errored with `Could not open Pine Editor` even when the editor was visibly open.

2. **Wrong editor index**: once `monacoEnv` was found, `FIND_MONACO` returned `getEditors()[0]` — the *oldest* mounted editor. With multiple editor tabs open, reads and writes went to a hidden background editor while the visible tab showed different code: `pine_set_source` reported success but the displayed buffer never changed, and `pine_smart_compile` compiled/saved a different tab than intended (a data-loss hazard for whichever script that tab held).

`FIND_MONACO` now prefers the visible container over DOM order, and picks the focused editor first, then a visible one, falling back to `editors[0]`.

Found and verified live against TradingView Desktop 3.3.0 (Electron 38.2.2): repro'd both failure modes with two editor tabs open, confirmed reads/writes/compile target the active tab after the fix.

### DIFF
--- a/src/core/pine.js
+++ b/src/core/pine.js
@@ -8,7 +8,14 @@ import { evaluate, evaluateAsync, getClient } from '../connection.js';
 // ── Monaco finder (injected into TV page) ──
 const FIND_MONACO = `
   (function findMonacoEditor() {
-    var container = document.querySelector('.monaco-editor.pine-editor-monaco');
+    // Panel close/reopen can leave a stale invisible container behind —
+    // prefer the visible instance (offsetParent set) over DOM order.
+    var candidates = document.querySelectorAll('.monaco-editor.pine-editor-monaco');
+    var container = null;
+    for (var ci = 0; ci < candidates.length; ci++) {
+      if (candidates[ci].offsetParent !== null) { container = candidates[ci]; break; }
+    }
+    if (!container) container = candidates[0] || null;
     if (!container) return null;
     var el = container;
     var fiberKey;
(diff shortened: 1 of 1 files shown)


shigechika:
Independently verified this fix end-to-end against a live TradingView Desktop session (macOS, Electron build) that already had the bug this PR fixes (2 `.monaco-editor.pine-editor-monaco` nodes in the DOM, one stale/detached — see repro details on #333).

**Before this patch**: `pine_get_source`/`pine_open` reliably failed after the first script in the session with `Could not open Pine Editor or Monaco not found in React fiber tree.`, or silently returned stale content from a previous script.

**After checking out this branch and restarting the MCP server**: opened 4 different saved scripts back-to-back and confirmed each `pine_open` + `pine_get_source` call returned the correct, distinct source (64 → 9 → 36 → 394 lines) — verified through both the standalone CLI (`tv pine open`/`tv pine get`) and the MCP tools (`pine_open`/`pine_get_source`). No failures, no stale content.

The `offsetParent !== null` container filter plus the `hasTextFocus()`/visible-DOM-node editor-index fix both look correct and match the maintainer's suggested approach on #333. This resolves #333 and #305 for me — recommend merging.

PR #346, opened 2026-07-11
Add MSIX-aware launch script that preserves package identity

CyberPS5:
## Problem

On current Windows builds, TradingView Desktop ships **only as an MSIX package** (installed under `C:\Program Files\WindowsApps\`). Both existing launch paths fail on such installs:

- **`scripts/launch_tv_debug.bat`** resolves the exe inside `WindowsApps` and runs it directly → **`Access is denied`** (Windows blocks direct execution of packaged binaries).
- **The `SETUP_GUIDE.md` copy-to-`%LOCALAPPDATA%` fallback** → the loose copy exits immediately with **`0x80000003` (STATUS_BREAKPOINT)** because it has no MSIX **package identity**. The `tv_launch` copy-fallback relies on this working, so it fails the same way.

Reproduced on `TradingView.Desktop 3.3.0.7992` / Electron 38, Windows 10 19045. The package also declares **no App Execution Alias**, so there's no alias to launch through either.

## Fix

Add `scripts/launch_tv_debug_msix.ps1`, which launches via **`Invoke-CommandInDesktopPackage`** — the supported mechanism that runs the *real* packaged exe **with full package identity** while still **forwarding `--remote-debugging-port`**. This is the only approach that produced a live CDP endpoint on the affected build.

The script:
- Resolves `PackageFamilyName`, `AppId`, and the exe path at runtime via `Get-AppxPackage` / the package manifest, so it survives version updates (no hardcoded `3.3.0.7992` path).
- Skips relaunch if CDP is already listening on the port.
- Polls `http://127.0.0.1:<port>/json/version` after launch and prints the WebSocket debugger URL, or errors out after ~60s.
- Takes an optional `-Port` parameter (default `9222`).

## Verification

```
> powershell -ExecutionPolicy Bypass -File scripts\launch_tv_debug_msix.ps1
Package : TradingView.Desktop_n534cwy3pjxzj
AppId   : TradingView.Desktop
Exe     : C:\Program Files\WindowsApps\TradingView.Desktop_3.3.0.7992_x64__n534cwy3pjxzj\TradingView.exe
Port    : 9222
CDP ready at http://127.0.0.1:9222
  Browser: Chrome/140.0.7339.133
```

`GET /json/version` confirmed reachable; the MCP server connected over CDP afterward.

## Notes

This is purely additive — it does not touch `launch_tv_debug.bat`. A natural follow-up (not included here) would be to have `tv_launch` prefer `Invoke-CommandInDesktopPackage` on MSIX installs, and to update `SETUP_GUIDE.md` Step 3 to note that the loose-copy fallback fails on builds that enforce package identity.

### DIFF
--- a/scripts/launch_tv_debug_msix.ps1
+++ b/scripts/launch_tv_debug_msix.ps1
@@ -1,0 +1,77 @@
+<#
+.SYNOPSIS
+    Launch TradingView Desktop (MSIX/Store install) with Chrome DevTools Protocol enabled.
+
+.DESCRIPTION
+    On modern Windows, TradingView ships as an MSIX package installed under
+    C:\Program Files\WindowsApps. Two common approaches FAIL on such builds:
+      * Running the exe directly from WindowsApps  -> "Access is denied"
+      * Copying the package elsewhere and running it -> exits with 0x80000003
+        (STATUS_BREAKPOINT) because the loose copy has no MSIX package identity.
+
+    The reliable method is Invoke-CommandInDesktopPackage, which launches the real
+    packaged exe WITH its package identity AND forwards command-line arguments.
+
+.PARAMETER Port
+    CDP port to expose. Default 9222.
+
+.EXAMPLE
+    powershell -ExecutionPolicy Bypass -File scripts\launch_tv_debug_msix.ps1
+    powershell -ExecutionPolicy Bypass -File scripts\launch_tv_debug_msix.ps1 -Port 9333
+#>
+param(
+    [int]$Port = 9222
+)
+
+$ErrorActionPreference = 'Stop'
+
+# --- Resolve the installed MSIX package (no admin required) ---
+$pkg = Get-AppxPackage -Name 'TradingView.Desktop' -ErrorAction SilentlyContinue
+if (-not $pkg) {
+    Write-Error "TradingView.Desktop MSIX package not found. Is TradingView Desktop installed?"
+    exit 1
+}
+
+$pfn = $pkg.PackageFamilyName
+$manifest = Get-AppxPackageManifest $pkg
+$app = $manifest.Package.Applications.Application | Select-Object -First 1
+$appId = $app.Id
+$exe = Join-Path $pkg.InstallLocation $app.Executable
+
+Write-Host "Package : $pfn"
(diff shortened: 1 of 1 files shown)


CyberPS5:
Verification update — this script has been my only launch path since I opened the PR, and it has held up under sustained real-world use.

**Environment** (same build as the original report):

- `TradingView.Desktop` **3.3.0.7992** (MSIX, resolved via `Get-AppxPackage`)
- CDP browser: `Chrome/140.0.7339.133` (Electron 38)
- Windows 10.0.19045

**What I ran:** the currently running TradingView process was started by `launch_tv_debug_msix.ps1` and has been up **44.9 hours on a single launch**, driving continuous CDP automation throughout — chart symbol/timeframe switching, `data_get_study_values`, Pine editor open + `pine_check` compiles, `indicator_search` / `indicator_add`, indicator-condition alert creation through the UI, and screenshot capture. No CDP drops, no relaunch needed, no elevation.

`tv_health_check` at time of writing:

```json
{
  "success": true,
  "cdp_connected": true,
  "api_available": true,
  "chart_symbol": "TVC:DJI",
  "chart_resolution": "1W"
}
```

The re-entrancy guard also does its job in practice: re-running the script while CDP is already listening exits `0` without touching the running instance, so it is safe to call from a wrapper without checking first.

**On the two failure modes in the PR body** — those remain as originally reported on this build; I have not re-run them since, so treat them as the July observations rather than fresh ones. `Invoke-CommandInDesktopPackage` sidesteps both, needs no admin rights, and copies nothing (versus the ~330MB local-copy fallback).

**One small offer, separate from this PR:** if you think it would be useful, I would be glad to add a short note to `SETUP_GUIDE.md` Step 3 mentioning this identity-preserving route alongside the existing `%LOCALAPPDATA%` copy fallback — so anyone on an MSIX install has a second option to reach for if the copy does not start on their particular build. Completely your call on whether that is worth the words, and I would keep it as its own docs-only PR per the one-change-per-PR guidance in CONTRIBUTING.

Thanks for maintaining this — the CDP bridge has been genuinely useful, and I am happy to adjust anything here to fit how you would prefer the Windows launch paths to be organised.

PR #355, opened 2026-07-15
Fix launch() crash on synchronous EPERM + bump chrome-remote-interface

rommdhoni:
## Summary
- Bump `chrome-remote-interface` to `^0.34.0`.
- Fix `launch()` in `src/core/health.js`: on some Windows MSIX setups, `spawn()` throws `EPERM` synchronously instead of emitting an async `error` event, which skipped the existing local-copy fallback entirely and crashed the whole launch. `_spawnDetached` now catches the synchronous throw and re-surfaces it as an `error` event so the fallback logic handles both cases uniformly.

## Test plan
- [x] `node --test tests/launch.test.js` - all 7 tests pass
- [x] Verified against a real Windows MSIX TradingView install: `launch()` now self-heals via the local-copy fallback and reaches a working CDP connection instead of throwing

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -9,7 +9,7 @@
       "version": "1.0.0",
       "dependencies": {
         "@modelcontextprotocol/sdk": "^1.12.1",
-        "chrome-remote-interface": "^0.33.2"
+        "chrome-remote-interface": "^0.34.0"
       },
       "bin": {
         "tv": "src/cli/index.js"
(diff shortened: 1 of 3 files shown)


Kx23-Hub:
Confirming the synchronous-EPERM case is real. On Windows 11 Home with the Microsoft Store build of TradingView Desktop 3.3.0, `spawn` on the WindowsApps exe throws `spawn EPERM` synchronously (not via the async `error` event), so `launch()` dies with `{"success": false, "error": "spawn EPERM"}` before the local-copy fallback ever runs. With the sync throw routed into the same failure path, the MSIX local-copy fallback kicks in and CDP binds fine on 9222. Wrapping it as an already-failed EventEmitter in `_spawnDetached` is a clean way to unify both cases.

🤖 Tested with [Claude Code](https://claude.com/claude-code)

PR #358, opened 2026-07-16
Fix MSIX package lookup to match non-Desktop-suffixed package IDs

PonchoJimmy:
`Get-AppxPackage -Name 'TradingView.Desktop'` misses installs published under a different package ID (e.g. `31178TradingViewInc.TradingView`), which some Microsoft Store / tvd-packages installs use. This widens the lookup to a wildcard match (`*TradingView*`) so it finds the app regardless of the publisher's exact package name.

Tested locally on Windows 11 against a `31178TradingViewInc.TradingView_3.3.0.0` install — `tv launch` now succeeds where it previously failed with "TradingView not found on win32."

### DIFF
--- a/src/core/health.js
+++ b/src/core/health.js
@@ -318,7 +318,7 @@ export async function launch({ port, kill_existing, _deps } = {}) {
     // MSIX/Windows Store install — InstallLocation is in WindowsApps, which is ACL-restricted
     // for normal `dir` enumeration but readable via Get-AppxPackage without elevation.
     try {
-      const ps = 'powershell -NoProfile -Command "(Get-AppxPackage -Name \'TradingView.Desktop\' -ErrorAction SilentlyContinue).InstallLocation"';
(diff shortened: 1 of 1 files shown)


Kx23-Hub:
Confirming this fixes a real-world install. On Windows 11 Home with the Microsoft Store build of TradingView Desktop 3.3.0, the package is named `31178TradingViewInc.TradingView` (full name `31178TradingViewInc.TradingView_3.3.0.0_x64__q4jpyh43s5mv6`), so the current `Get-AppxPackage -Name 'TradingView.Desktop'` lookup returns nothing and `launch` fails with "TradingView not found". With the `*TradingView*` wildcard, detection works and CDP comes up.

One small suggestion: if more than one package ever matches, `.InstallLocation` comes back as multiple lines, and appending `\TradingView.exe` to the joined string would produce a bad path. Splitting on newlines and checking `existsSync` per candidate would make it robust.

🤖 Tested with [Claude Code](https://claude.com/claude-code)

Issue #143, opened 2026-05-11
data_get_study_values deduplicates studies by name — returns only one entry for multiple instances with the same name


JDuckling:
## Description

When multiple study instances on the chart share the same `name` (e.g. two `Moving Average Exponential` studies — one at length=50, one at length=200), `data_get_study_values` returns fewer entries than `chart_get_state` shows. Specifically, the `study_count` field in the response is less than the actual number of studies, and only one of the same-name studies returns values (usually the first by add-order).

The two studies have distinct `entity_id`s in `chart_get_state` output, so the deduplication is happening downstream — likely in the values-aggregation logic.

## Steps to reproduce

1. Add two Moving Average Exponential studies with different lengths (see issue #02 for the input-handling workaround; manually configure if needed):
   - EMA-50 close → entity_id "A"
   - EMA-200 close → entity_id "B"
2. `chart_get_state` confirms 2 separate studies in `studies[]` array with the same `name: "Moving Average Exponential"`.
3. Call `data_get_study_values`.

## Expected behavior

Response includes 2 entries for `Moving Average Exponential`, with distinct values:
```json
{
  "studies": [
    {"name": "Moving Average Exponential", "values": {"EMA": "76031"}},  // EMA-50
    {"name": "Moving Average Exponential", "values": {"EMA": "82025"}}   // EMA-200
  ]
}
```

## Actual behavior (pre-workaround)

Earlier observation (universal-core plan 2026-05-10): `data_get_study_values` returned `study_count: 4` while `chart_get_state` listed 5 studies. The missing one was the second instance of `Moving Average Exponential`.

After today (2026-05-11) — newer observation suggests both EMAs now do return separately. So the deduplication might be partial or fixed in some path. Needs re-verification on a fresh chart with two identical-name studies.

```json
// Current observed (good):
{
  "name": "Moving Average Exponential",
  "values": {"EMA": "76 031"}
},
{
  "name": "Moving Average Exponential",
  "values": {"EMA": "82 025"}
}
```

## Environment

- macOS: Darwin 25.4.0
- TradingView Desktop: stable channel
- TV plan: Essential
- MCP fork: `JDuckling/tradingview-mcp` (likely upstream too)
- Date originally observed: 2026-05-10. Re-tested 2026-05-11 — appears to work now but worth a regression test on the maintainer side.

## Workaround

[...]

PR #363, opened 2026-07-19
Strip Pine source blob from data_get_study_values inputs

frkdkl:
## Problem

`data_get_study_values` returns each study's full `inputs()` object. For Pine studies that carries the script's **identity** alongside its configuration: the obfuscated source (`text`, multi-KB), `pineId`, `pineVersion`, `pineFeatures`, and the internal `__fast_calc` / `__profile` flags. None of it describes how the study is configured.

This matters because the tool's output goes straight into an LLM context window. On a chart carrying several Pine toolkits, the blob crowds out the values the call was made for.

## What this does

Drops the six identity/metadata keys, unwraps the `{v, f, t}` envelope so the remaining inputs stay readable, and caps string values at 500 chars as a safety net for unknown blobs.

`pineId` deserves a note, because it is the one that looks useful and isn't: it is **identical across every instance of the same script**, so it cannot help tell instances apart. What does that — and what #143 asked for — is `in_0..in_N`, which stay.

The 500-char cap is the same threshold `getIndicator()` uses in this file (`src/core/data.js:203`). The name list alone would let a Pine script with a long text input slip through; the cap alone would keep the small boolean flags. Both are needed, which is why this doesn't simply reuse `getIndicator()`'s filter — that one operates on the array from `getInputValues()`, an API the data-source objects reached here don't expose.

## Verification

Live against TradingView Desktop 3.3.0 (Electron 38.2.2, CDP 9222), macOS.

`JSON.stringify` length of the `inputs` payload for the same chart state:

| | chars |
|---|---|
| raw | 1373 |
| name filter only | 143 |
| **this PR** | **74** |

The Pine study itself went 1352 → 53. A non-Pine data source on the same chart is untouched at 21 chars, confirming the filter doesn't reach past Pine metadata.

Values themselves are unchanged. Verified across a custom Pine study (4 configured inputs, all preserved: `{in_0: true, in_1: 163.988, in_2: true, in_3: 162.2}`), built-in RSI and built-in MACD — all returned complete and identical before and after.

`npm run lint` clean and `npm run test:unit` 152/152 on the branch.

## Note on tests

No unit test added: the changed code is page-context JS inside the `evaluate()` template string, so it is only reachable through a live CDP session, and extracting it into a Node-testable function would mean restructuring well beyond this fix. Happy to add one if you would like it covered a particular way.

### DIFF
--- a/src/core/data.js
+++ b/src/core/data.js
@@ -524,19 +524,26 @@ export async function getStudyValues() {
           // (e.g. two EMAs with different lengths) are distinguishable (#143).
           var id = null;
           try { id = s.id ? s.id() : null; } catch(e) {}
-          // Pine studies stash their obfuscated source in inputs().text — multi-KB
-          // per study, and it says nothing about how the study is configured. Drop
-          // it and the internal flags, and unwrap the {v,f,t} envelope so the
-          // inputs stay readable while still distinguishing instances (#143).
+          // A Pine study's inputs() carries its identity alongside its configuration:
+          // the obfuscated source (text, multi-KB), the script id and version, the
+          // feature list and the internal calc flags. None of it says how the study
+          // is configured, and pineId is identical across every instance of a script
+          // — so it cannot help tell instances apart either (#143). Drop that set,
+          // keep the in_* values that do, and unwrap the {v,f,t} envelope.
+          var META = ['text', 'pineId', 'pineVersion', 'pineFeatures', '__fast_calc', '__profile'];
           var inputs = null;
           try {
             var ip = s.inputs ? s.inputs() : null;
             if (ip) {
               var lean = {};
               for (var k in ip) {
-                if (k === 'text' || k === 'pineFeatures' || k === '__fast_calc' || k === '__profile') continue;
+                if (META.indexOf(k) !== -1) continue;
                 var iv = ip[k];
-                lean[k] = (iv && typeof iv === 'object' && 'v' in iv) ? iv.v : iv;
+                var val = (iv && typeof iv === 'object' && 'v' in iv) ? iv.v : iv;
+                // Safety net for unknown blobs: getIndicator() drops strings over
+                // 500 chars (src/core/data.js:203) — same threshold here.
+                if (typeof val === 'string' && val.length > 500) continue;
+                lean[k] = val;
               }
               if (Object.keys(lean).length) inputs = lean;
             }
--- a/src/core/data.js
+++ b/src/core/data.js
@@ -524,8 +524,23 @@ export async function getStudyValues() {
           // (e.g. two EMAs with different lengths) are distinguishable (#143).
           var id = null;
           try { id = s.id ? s.id() : null; } catch(e) {}
+          // Pine studies stash their obfuscated source in inputs().text — multi-KB
+          // per study, and it says nothing about how the study is configured. Drop
+          // it and the internal flags, and unwrap the {v,f,t} envelope so the
+          // inputs stay readable while still distinguishing instances (#143).
           var inputs = null;
-          try { var ip = s.inputs ? s.inputs() : null; if (ip && Object.keys(ip).length) inputs = ip; } catch(e) {}
+          try {
+            var ip = s.inputs ? s.inputs() : null;
+            if (ip) {
+              var lean = {};
+              for (var k in ip) {
+                if (k === 'text' || k === 'pineFeatures' || k === '__fast_calc' || k === '__profile') continue;
+                var iv = ip[k];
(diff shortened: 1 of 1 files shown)


L3G1T50:
Hit this today and went looking for the fix before writing one — this PR already covers it, so here is independent corroboration on a heavier chart rather than a competing change.

I ran your exact filter (same six `META` keys, same `{v,f,t}` unwrap, same 500-char cap) in page context against a 13-study chart carrying two protected Pine toolkits, measuring `JSON.stringify(inputs).length` for the studies `getStudyValues()` actually emits:

| | chars |
| --- | --- |
| raw | **21,978** |
| name filter only | 1,716 |
| **this PR** | **552** |

Per study:

| study | raw | filtered | `text` alone |
| --- | --- | --- | --- |
| PDH/PDL Liquidity Reversal | 14,694 | 436 | 13,058 |
| ICT GODMODE INSTITUTIONAL AI | 7,284 | 116 | 6,574 |

~40× overall, against the ~18.5× on your chart — protected indicators make it considerably worse than the case you measured. The two `text` blobs alone are 19,632 chars.

Worth noting the `{v,f,t}` unwrap carries real weight on its own: 1,716 → 552 after the name filter has already run, so roughly two thirds of what survives the key list is envelope. Dropping the six keys without unwrapping would leave most of the remaining bloat in place.

One angle not in the description, which I think strengthens the case: **the repo already documents this behaviour as if it were implemented.**

- `README.md:325` — "| Indicator inputs | Encrypted/encoded blobs auto-filtered |"
- `CLAUDE.md:95` — "Avoid calling `data_get_indicator` on protected/encrypted indicators — their inputs are encoded blobs. **Use `data_get_study_values` instead** for current values."

So the context-management table promises filtering that does not happen, and the agent-facing guidance routes you to `data_get_study_values` specifically to *avoid* encoded blobs — straight into the path that emits them. Following CLAUDE.md as written is what produced the 21,978 chars above. This PR makes both statements true rather than requiring a docs change.

For what it is worth I think the no-unit-test call is right. The filter lives inside the `evaluate()` template string, and pulling it out into a Node-testable function would restructure a good deal more than the fix warrants. The before/after character counts are the meaningful evidence here, and they reproduce.

Environment: TradingView Desktop 3.3.0.7992 (Electron 38.2.2), Windows 11, Node 24.14.1, CDP 9222.

chasewhughes:
Independent verification at current main (`c05b8f5`). Applies cleanly, still `MERGEABLE`.

**Environment:** macOS 26.6 (arm64), Node v22.16.0, TradingView Desktop 3.3.0, CDP 9222.

### Measured before/after on the same chart

IWM daily, 3 studies loaded (Volume, SMA, Relative Strength Index), measuring `JSON.stringify()` of the full `getStudyValues()` return:

| | Payload |
|---|---|
| main @ `c05b8f5` | **6,183 bytes** |
| with this PR | **719 bytes** |

88% reduction on a 3-study chart. The `text` blob alone is ~2.5 KB per Pine study, so the saving scales with study count — on a chart carrying 6–8 indicators this is the difference between a usable tool call and one that dominates the context window. It matters most for multi-symbol scans, where the tool is called once per symbol.

### Nothing lost

All `values` are byte-identical before and after. The `in_*` inputs that distinguish instances (#143) are preserved, and unwrapping the `{v,f,t}` envelope makes them more readable rather than less:

```
before:  "in_1": { "v": 50, "f": true, "t": "integer" }
after:   "in_1": 50
```

Sample of the full post-PR output for the three studies:

```
Volume    inputs {"length":20,"col_prev_close":false}
          values {"Volume":"14.71 M"}
SMA       inputs {"in_0":"close","in_1":50,"in_2":200,...}
          values {"10":"298.50","20":"295.77","50":"294.52","100":"282.69","200":"268.01"}
RSI       inputs {"in_0":14,"in_1":"close","in_3":"SMA","in_4":14,...}
          values {"RSI":"61.42","RSI-based MA":"54.45"}
```

`npm run lint` → 0 errors (4 pre-existing warnings). `npm run test:unit` → 152/152 pass.

### One consistency point in this PR's favour

`getIndicator()` already drops the same blob — the `id === 'text'` / `length > 500` filter in `src/core/data.js`. Before this PR the two tools disagreed about whether a multi-KB obfuscated Pine source counts as "an input". After it, they agree.

frkdkl:
Author here — thanks to both of you, that's more verification than I could have produced myself.

Status from my side: rebased onto `c05b8f5`, `npm run test:unit` 152/152 locally. Happy to rebase again, split the commits, or narrow the filter list if any of that helps review.

PR #366, opened 2026-07-19
fix(launch): detect MSIX TradingView by wildcard, not exact package name

iamdrsid1:
## Problem

On Windows, `tv launch` (and the `tv_launch` MCP tool) fail with **"TradingView not found on win32"** even when TradingView Desktop is installed as an MSIX/Store package.

The MSIX detection in `src/core/health.js` queries:

```powershell
(Get-AppxPackage -Name 'TradingView.Desktop').InstallLocation
```

But the installed package name is **publisher-prefixed** — e.g. `31178TradingViewInc.TradingView` — so the exact `-Name 'TradingView.Desktop'` match returns nothing. Detection then falls through to the legacy install paths (`%LOCALAPPDATA%\TradingView`, `%PROGRAMFILES%\TradingView`, …), none of which exist for a Store install, and throws.

Real install observed:
```
C:\Program Files\WindowsApps\31178TradingViewInc.TradingView_3.3.0.0_x64__q4jpyh43s5mv6
```

## Fix

Match on a `*TradingView*` wildcard and select the first install whose folder actually contains `TradingView.exe`:

```powershell
Get-AppxPackage *TradingView* | ForEach-Object { $_.InstallLocation } |
  Where-Object { $_ -and (Test-Path (Join-Path $_ 'TradingView.exe')) } |
  Select-Object -First 1
```

The `Test-Path` guard avoids picking up an unrelated `*TradingView*` package that has no launcher. The existing WindowsApps → `%LOCALAPPDATA%` local-copy fallback (for the "Access is denied" case) is unchanged.

## Verification

- Resolves the real MSIX install via the exact `execSync` code path (v3.3.0.0).
- `tests/launch.test.js` → **7/7 pass** (the mock keys on the `Get-AppxPackage` substring, so it stays compatible).
- ESLint clean.

### DIFF
--- a/src/core/health.js
+++ b/src/core/health.js
@@ -317,8 +317,11 @@ export async function launch({ port, kill_existing, _deps } = {}) {
   if (!tvPath && platform === 'win32') {
     // MSIX/Windows Store install — InstallLocation is in WindowsApps, which is ACL-restricted
     // for normal `dir` enumeration but readable via Get-AppxPackage without elevation.
+    // The package Name is publisher-prefixed (e.g. `31178TradingViewInc.TradingView`),
+    // not the bare `TradingView.Desktop`, so match on a `*TradingView*` wildcard and pick
+    // the first install whose folder actually contains TradingView.exe.
     try {
-      const ps = 'powershell -NoProfile -Command "(Get-AppxPackage -Name \'TradingView.Desktop\' -ErrorAction SilentlyContinue).InstallLocation"';
+      const ps = 'powershell -NoProfile -Command "Get-AppxPackage *TradingView* | ForEach-Object { $_.InstallLocation } | Where-Object { $_ -and (Test-Path (Join-Path $_ \'TradingView.exe\')) } | Select-Object -First 1"';
       const installDir = deps.execSync(ps, { timeout: 5000 }).toString().trim();
       if (installDir) {
         const candidate = `${installDir}\\TradingView.exe`;


AppCryptoLab:
I hit this exact bug on a Microsoft Store install (Windows 11, TradingView Desktop 3.3.0, package identity `31178TradingViewInc.TradingView`) and independently arrived at the same fix before finding this PR. Confirming it works, and offering a regression test rather than opening a duplicate.

**Why the current test suite didn't catch this**

`tests/launch.test.js` passes against the unpatched code. The `execSync` mock returns a `TradingView.Desktop_*` install location for *any* `Get-AppxPackage` command:

```js
execSync: (cmd) => {
  if (cmd.includes('Get-AppxPackage')) {
    return 'C:\\Program Files\\WindowsApps\\TradingView.Desktop_3.1.0.7818_x64__n534cwy3pjxzj\n';
  }
  ...
```

So the exact `-Name` filter is never actually exercised — the mock answers as if it matched. The bug is invisible to the suite, which is presumably why several duplicate PRs for it exist.

**A test that models PowerShell's real semantics**

The fixed-name query returns empty; only the wildcard matches:

```js
const PREFIXED_DIR = 'C:\\Program Files\\WindowsApps\\31178TradingViewInc.TradingView_3.3.0.0_x64__q4jpyh43s5mv6';
const PREFIXED_EXE = `${PREFIXED_DIR}\\TradingView.exe`;

it('finds a publisher-prefixed AppX identity that a fixed-name query misses', async () => {
  const state = { spawned: [], cdpUp: false, queries: [] };
  const deps = {
    existsSync: (p) => p === PREFIXED_EXE,
    execSync: (cmd) => {
      if (cmd.includes('taskkill')) return '';
      if (cmd.includes('Get-AppxPackage')) {
        state.queries.push(cmd);
        return cmd.includes('*TradingView*') ? `${PREFIXED_DIR}\n` : '';
      }
      throw new Error(`unexpected execSync: ${cmd}`);
    },
    spawn: (exe) => { state.spawned.push(exe); state.cdpUp = true; return mockChild(); },
    cpSync: () => { throw new Error('should not copy'); },
    rmSync: () => {},
    readdirSync: () => [],
    delay: async () => {},
    probeCdp: async () => (state.cdpUp ? CDP_VERSION : null),
  };
  const result = await launch({ _deps: deps });
  assert.equal(result.success, true);
  assert.equal(result.binary, PREFIXED_EXE);
  assert.ok(state.queries.length > 0, 'should have queried Get-AppxPackage');
});
```

Fails on `main`, passes with this PR applied. It matches on `*TradingView*` being present in the command, so it is agnostic to whether the `Test-Path` filtering happens in PowerShell (as here) or in JS — it won't lock in either implementation.

**This PR fixes one of two independent MSIX bugs**

Running that suite against this branch gives 8/9, with one unrelated failure:

```
✔ finds a publisher-prefixed AppX identity that a fixed-name query misses
✖ falls back to a local copy when the direct spawn throws synchronously
```

The remaining failure is a separate bug — `deps.spawn` can throw `EPERM` synchronously for WindowsApps binaries, which escapes `launch()` before the local-copy fallback runs. #337 fixes that half. The two PRs are complementary rather than competing; together they take the suite to 9/9.

Both tests are on [`3675d29`](https://github.com/AppCryptoLab/tradingview-mcp/commit/3675d294afa9689e137d554545f4f158f672de3a) if it's useful to cherry-pick. Happy to open it as a tests-only PR once a fix lands, though it would sit red on `main` until then. Feel free to take the code as-is, no attribution needed.

Danielworkapp:
Verified on a Microsoft Store install.

**Environment:** Windows 11 Pro 10.0.26200, TradingView Desktop 3.4.0 from the Microsoft Store (package `31178TradingViewInc.TradingView`, version 3.4.0.0), Node 24.19.0.

**`main` (c05b8f5)**, calling `launch()` from `src/core/health.js` directly:

```
threw: TradingView not found on win32. Searched: %LOCALAPPDATA%\TradingView\TradingView.exe, C:\Program Files\TradingView\TradingView.exe, C:\Program Files (x86)\TradingView\TradingView.exe. ...
```

**This PR (85f1653)**, same call:

```json
{"success":true,"platform":"win32","binary":"C:\\Program Files\\WindowsApps\\31178TradingViewInc.TradingView_3.4.0.0_x64__q4jpyh43s5mv6\\TradingView.exe","cdp_port":9222,"cdp_url":"http://127.0.0.1:9222","browser":"Chrome/146.0.7680.216", ...}
```

The debug port opened in about 7s, launched directly from `WindowsApps`. This build didn't need the local-copy fallback.

`npm run lint` is clean. `npm run test:unit` on Windows: 123/125, and the same two tests fail on `main` (Pine compile and the source audit), so neither failure comes from this PR.

I also ran #485 (1ca0772) the same way and got the same result: launched from `WindowsApps`, debug port up in about 6s, 127/129 unit tests, and its 4 new launch tests pass.

PR #367, opened 2026-07-19
Honor explicitly owned TradingView CDP targets

BeeSting:
## Summary

- honor `TRADINGVIEW_MCP_TARGET_ID` when a CLI process opens its CDP connection
- select the explicitly owned chart tab instead of the first TradingView tab returned by Chrome
- accept Ticker Alpha's `TRADINGVIEW_CDP_PORT` compatibility variable and cover both behaviors with unit tests

## Root cause

Ticker Alpha opens a dedicated CDP chart target and passes its ID to every TradingView MCP CLI command. The MCP connection layer ignored that environment binding and selected the first TradingView chart target. With an existing chart tab ahead of the newly opened target, Ticker Alpha's owned-target attestation failed closed.

## Validation

- `npm run test:unit` — 155 passed, 0 failed
- `npm run lint` — 0 errors; 4 pre-existing warnings
- production canary reproduced the failure before this fix: the owned ABBV target differed from the first existing IREN target
- full live E2E remains pending because this workstation has no TradingView CDP session on port 9222; the PR is intentionally draft until it is installed and canaried on the Mac Mini

### DIFF
--- a/package.json
+++ b/package.json
@@ -17,9 +17,9 @@
     "lint": "eslint src/",
     "test": "node --test tests/e2e.test.js tests/pine_analyze.test.js",
     "test:e2e": "node --test tests/e2e.test.js",
-    "test:unit": "node --test tests/pine_analyze.test.js tests/cli.test.js tests/sanitization.test.js tests/replay.test.js tests/launch.test.js tests/chart_indicator.test.js tests/chart_history.test.js tests/update.test.js",
+    "test:unit": "node --test tests/pine_analyze.test.js tests/cli.test.js tests/sanitization.test.js tests/replay.test.js tests/launch.test.js tests/chart_indicator.test.js tests/chart_history.test.js tests/update.test.js tests/connection.test.js",
     "test:cli": "node --test tests/cli.test.js",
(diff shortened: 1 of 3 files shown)


fantasy5797:
I tested this branch locally and proposed a focused follow-up on top of it: https://github.com/BeeSting/tradingview-mcp/pull/1

The increment adds stable saved-chart ownership (`TRADINGVIEW_MCP_CHART_ID`, plus compatibility aliases) so selection survives ephemeral CDP target-id changes. It also preserves explicit `tab_switch` target overrides and fails closed when target/chart selectors disagree.

Verification: 160 unit tests passed, lint had 0 errors, and live TradingView Desktop CDP selected two distinct saved chart ids with the expected target/symbol/resolution.

tradesdontlie:
This looks awesome thank you I will test this ASAP

codeCraft-Ritik (commented):
Nice fix! Respecting TRADINGVIEW_MCP_TARGET_ID when establishing the CDP connection makes the target selection deterministic and aligns the CLI behavior with the ownership model expected by Ticker Alpha. The added compatibility support and tests are a nice touch as well.

PR #368, opened 2026-07-20
Fix Windows CLI exit crash, MSIX launch EPERM fallback, and a test path bug

jelangbajapana:
Three Windows-specific fixes, verified on a real MSIX TradingView install. Unit suite: 160/160 pass.

1. CLI crashes on exit (Windows) — src/cli/router.js. Commands printed correct output then crashed with "Assertion failed: !(handle->flags & UV_HANDLE_CLOSING), file src\win\async.c" and a nonzero exit code (e.g. `tv pine check`). The router called process.exit() while async handles were still closing — undici (Node's fetch) and the CDP WebSocket — racing libuv teardown. It now sets process.exitCode, closes the CDP connection, and lets t
2. he event loop drain naturally, with an unref'd safety timer.
3. MSIX launch fails on synchronous spawn EPERM — src/core/health.js. launch() only fell back to the local-copy workaround when the WindowsApps spawn failed via an async EACCES 'error' event. On some machines spawn() throws EPERM synchronously, which escaped launch() entirely so the documented fallback never ran (tv launch returned {"error":"spawn EPERM"}). The initial spawn is now guarded; a throw on an MSIX path routes into the local-copy fallback, matching how async EACCES is handled.
4. Windows path bug in tests — tests/sanitization.test.js. Used new URL(...).pathname, which yields "/C:/..." on Windows and made readdirSync resolve to "C:\C:\Users\...". Switched to fileURLToPath().
Added a launch.test.js case for the synchronous-EPERM failure mode. Full unit suite passes on Windows (was failing before).

### DIFF
--- a/src/cli/router.js
+++ b/src/cli/router.js
@@ -3,6 +3,25 @@
  * Zero dependencies — uses only Node.js built-ins.
  */
 import { parseArgs } from 'node:util';
+import { disconnect } from '../connection.js';
+
+/**
+ * Exit the CLI cleanly. Calling process.exit() abruptly races libuv's async
+ * handle teardown on Windows — undici (Node's fetch, used by `pine check`) and
+ * the CDP WebSocket both trip the assertion
+ *   Assertion failed: !(handle->flags & UV_HANDLE_CLOSING), file src\win\async.c
+ * crashing with a nonzero code even after a successful command. Instead we set
+ * the exit code, close the CDP connection, and let the event loop drain
+ * naturally: undici unrefs its keep-alive sockets and CDP's socket closes, so
+ * the process exits on its own with no race. An unref'd timer force-exits only
+ * if some handle genuinely hangs — it never delays a clean exit.
+ */
+async function exit(code) {
+  process.exitCode = code;
+  try { await disconnect(); } catch { /* already closed */ }
+  const t = setTimeout(() => process.exit(code), 3000);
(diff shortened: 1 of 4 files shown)


jelangbajapana:
Two further commits were added after the original description, both from testing on a real MSIX install where `tv launch` reported success but left no usable session.

**4. CDP was only probed once** — `src/core/health.js`
The app binds the debug port, its GPU process then crash-loops with an access violation (`exit_code=-2147483645`), and the app restarts itself without the debug flag. Processes stay alive but nothing listens on the port. `launch()` had already seen its single successful probe and reported success, handing back a dead endpoint. `_waitForLiveCdp` now re-probes after a grace period and only reports success if the endpoint is still there.

**5. No recovery from those crash loops** — `src/core/health.js`
A final retry respawns with `--disable-gpu`, `--disable-gpu-sandbox`, `--disable-software-rasterizer` and `--no-sandbox`. Both parts are needed: disabling the GPU stops the crash loop that kills the app, and disabling the sandbox stops the renderer crashing to an error page (`crashReason: crashed`) even once the app survives. Launches that need it report `crash_fallback: true`.

Verified end to end: `tv launch` returns `crash_fallback: true`, the session persists, and `tv_health_check` reports `api_available: true` against a live chart (OANDA:XAUUSD, 15m).

Unit suite is now 163/163 on Windows.

PR #391, opened 2026-07-25
fix: run update-check git commands with explicit repo cwd

deepfrequencyservice:
## Summary
- \`checkForUpdate()\` in \`src/core/health.js\` shelled out to \`git\` without passing an explicit \`cwd\`, so it inherited the Node process's current working directory instead of the MCP server's own repo directory.
- When the server is launched from a different working directory (e.g. via an MCP client config that starts the process elsewhere), \`tv_health_check\` reported the wrong \`local_commit\`, which could make \`update_available\` incorrect and confuse \`tv_update\`.
- Fix: run the update-check git commands with the server's own repo path as \`cwd\` explicitly.

## Test plan
- [x] Ran \`tv_health_check\` and \`tv_update\` against a checkout launched from an unrelated working directory; \`local_commit\` now reflects the server's actual repo state.
- [x] \`tv_update\` still correctly refuses to fast-forward when the local branch has commits not on \`origin/main\` (verified against a repo with an unpushed local commit).

### DIFF
--- a/src/core/health.js
+++ b/src/core/health.js
@@ -5,6 +5,10 @@ import { getClient, getTargetInfo, evaluate, CDP_HOST, CDP_PORT } from '../conne
 import { existsSync, cpSync, rmSync, readdirSync } from 'fs';
 import { execSync, spawn } from 'child_process';
 import { dirname, basename, join } from 'path';
+import { fileURLToPath } from 'url';
+
+// src/core/health.js -> repo root, independent of process.cwd()
+const REPO_ROOT = dirname(dirname(dirname(fileURLToPath(import.meta.url))));
 
 // Best-effort git-pull update check: compare local HEAD to origin's default
 // branch on GitHub. Never throws — returns null on any failure (offline,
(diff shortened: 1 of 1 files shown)


L3G1T50:
Heads-up: this overlaps with #456 (mine), which fixes a different bug in the same function. `checkForUpdate` reports an update whenever local `HEAD` differs from origin's, even when local is *ahead*. The two PRs conflict in `src/core/health.js` (checked with `git merge-tree`), so whichever one lands second will need a rebase.

That rebase needs one more change. #456 adds a `git` helper inside `checkForUpdate` and passes it to `classifyUpdate()`, which runs `git merge-base --is-ancestor <sha> HEAD` and `git rev-list --count <sha>..HEAD`. That helper has no `cwd`, so it has the same problem this PR fixes. If the server is started from another folder, the ancestor check fails and the update check reports `behind: true` again, so the phantom update comes back.

I checked this with local `HEAD` one commit ahead of the remote SHA:
- `classifyUpdate` with git run in the repo returns `{"behind":false,"ahead":1}`.
- The same call with git run in the temp folder returns `{"behind":true,"ahead":0}`.

So whichever PR merges second should also give that helper `cwd: REPO_ROOT`.

For reference, my fork has both fixes combined: one `repoGit()` helper runs `rev-parse HEAD` and `config --get remote.origin.url`, and it is also the helper passed to `classifyUpdate()`. A test runs it from the temp folder: L3G1T50/tradingview-mcp@19841c22134448462d055c7cc4a15a26088e1b6b

L3G1T50:
Upstream CI has never created a workflow run for this PR, so no run is waiting for approval. The workflow triggers on `opened`, `synchronize` and `reopened`, so a run would need a new push or a close/reopen.

I ran the same workflow in my fork against this PR's merge result, `refs/pull/391/merge` (`10464b5`, which is `3f792f0` merged into the current `main`, `c05b8f5`):

https://github.com/L3G1T50/tradingview-mcp/actions/runs/35157601667 (`success`)

The run used a throwaway branch: `10464b5` plus one commit that only adds `'ci/**'` to the `push` trigger in `.github/workflows/ci.yml`, so the workflow would run on push. The jobs and steps are unchanged, and no other file differs from the merge result.

- `npm ci`: passed.
- `npm run test:unit`: 152/152 passed.

This PR doesn't add tests, and no existing unit test calls `checkForUpdate()`, so the run shows the change doesn't break lint or the existing suite. It doesn't exercise the `cwd` fix itself; that still rests on the manual checks in the test plan.

PR #413, opened 2026-07-30
Complete DI refactor for remaining drawing.js functions

rzkarsyad:
## Summary

`drawShape` already resolves its dependencies through `_resolve(_deps)`, but the other four exported functions in `src/core/drawing.js` still call the module-level `_evaluate` / `_getChartApi` imports directly. That makes them the only functions in the module that can't be unit tested without a live CDP connection.

This applies the existing `_resolve(_deps)` pattern to the remaining four:

- `listDrawings`
- `getProperties`
- `removeOne`
- `clearAll`

## Behavior

No functional change. `_resolve()` falls back to the real `connection.js` implementations when no deps are passed, and every current caller passes none:

- `src/tools/drawing.js` — `core.listDrawings()`, `core.clearAll()`, `core.removeOne({ entity_id })`, `core.getProperties({ entity_id })`
- `src/cli/commands/drawing.js` — same call shapes

`listDrawings` and `clearAll` take deps as a positional arg (they have no other parameters); `getProperties` and `removeOne` take `_deps` in their existing options object. This matches how `drawShape` already does it.

## Testing

`node --test` across the offline suites — 152 passing, 0 failing. This includes `tests/sanitization.test.js`, which audits `drawing.js` for unsafe interpolation and still passes.

### DIFF
--- a/src/core/drawing.js
+++ b/src/core/drawing.js
@@ -44,9 +44,10 @@ export async function drawShape({ shape, point, point2, overrides: overridesRaw,
   return { success: true, shape, entity_id: result?.entity_id };
 }
 
-export async function listDrawings() {
-  const apiPath = await _getChartApi();
-  const shapes = await _evaluate(`
+export async function listDrawings(_deps) {
+  const { evaluate, getChartApi } = _resolve(_deps);
+  const apiPath = await getChartApi();
+  const shapes = await evaluate(`
     (function() {
       var api = ${apiPath};
       var all = api.getAllShapes();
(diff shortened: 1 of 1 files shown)


bangboomben:
Can confirm this one — I hit the same `ReferenceError` on `draw_list` / `draw_get_properties` / `draw_remove_one` / `draw_clear` against TV Desktop and ended up making an almost identical change locally before finding this PR.

Two things that might be useful:

**Regression tests.** I wrote 5 tests against the existing `mockDeps` helper in `tests/sanitization.test.js` that cover all four functions (injected-deps happy path, `safeString` escaping of `entity_id` in `getProperties`, and the shape-not-found rejection). They'd lock this in so it can't silently regress — the bug was invisible to the current suite because only `drawShape` was exercised. Happy to open them as a follow-up PR once this lands, or you're welcome to lift them straight into this branch, whichever the maintainers prefer.

**Heads-up if you do add tests here:** `tests/sanitization.test.js` currently fails on Windows before it gets to any new test, because the source-audit block resolves `CORE_DIR` via `new URL('../src/core/', import.meta.url).pathname`, which yields `/C:/...` and blows up `readdirSync`. #394 fixes exactly that with `fileURLToPath`, so test additions here effectively depend on it.

**Minor consistency note:** this makes `listDrawings(_deps)` and `clearAll(_deps)` take deps positionally, while `getProperties`/`removeOne`/`drawShape` destructure `{ _deps }` from an options object. No runtime impact — both call sites (`src/tools/drawing.js`, `src/cli/commands/drawing.js`) invoke with no arguments and `_resolve` falls back to the module imports either way. But `listDrawings({ _deps } = {})` would keep one calling convention across the module and let tests use the same shape everywhere.

PR #415, opened 2026-08-01
fix: make Pine script creation and saving reliable

huof6829:
## Summary

This PR fixes several reliability and safety issues in the Pine Script workflow, including script creation, editor updates, first-time saving, persistence verification, and script identity handling.

It also disables screenshot-related MCP operations so that the server exposes only structured TradingView interfaces.

### 1. `pine_new` did not create a real TradingView script

Previously, `pine_new` only replaced the current Monaco editor source with a template.

This caused several problems:

- no new TradingView script identity was created
- the editor could remain attached to the previously opened saved script
- a later save could overwrite the wrong script
- the result looked like a newly created script even though only the editor text had changed

### 2. `pine_save` could open the wrong context menu

The previous implementation selected a context-menu trigger based on vertical position.

The Pine Editor and Strategy Tester tabs can appear at the same Y coordinate, so the implementation could accidentally open the Strategy Tester menu instead of the Pine Editor menu.

### 3. First-time save dialogs were not handled

Saving a new TradingView script opens a `Save script` dialog that requires a script name.

Previously, the `name` argument was only used for verification. It was not entered into the dialog, so first-time saves could time out even when a name was supplied.

### 4. Save success did not guarantee persistence

Dispatching a keyboard shortcut or clicking a Save control does not prove that TradingView persisted the current editor source.

The previous implementation could report success without verifying the saved server-side source.

### 5. Monaco updates were not always registered as dirty changes

Directly calling `model.setValue()` could update the visible source without reliably registering the edit in TradingView's editor state.

This could leave the Save command disabled even though the displayed source had changed.

### 6. A closed Pine Editor could not be reopened reliably

`ensurePineEditorOpen()` attempted to activate the internal bottom-widget API and then called `HTMLElement.click()` on the Pine toolbar button.

In current TradingView Desktop builds, the Pine button can ignore synthetic DOM clicks. The internal bottom-widget methods can also exist while doing nothing for the right-side Pine panel. This caused structured Pine operations to fail with `Could not open Pine Editor.` even though the input and MCP configuration were valid.

A related recovery issue occurred when TradingView marked a blank `Untitled script` as dirty: `pine_new` refused to replace the empty buffer, leaving no structured way to recover.

### Genuine script creation

`pine_new` now uses TradingView's real UI workflow:

```text
Create new → Indicator / Strategy / Library
```

This creates a genuine TradingView script identity instead of replacing the source of the currently attached script.

Additional safeguards:

- refuses to create a new script when the current editor contains non-empty unsaved content
- allows a blank or whitespace-only dirty `Untitled script` buffer to be replaced for recovery
- waits for the Monaco editor to remount
- verifies that the generated source contains the requested declaration: `indicator(...)`, `strategy(...)`, or `library(...)`
- temporarily moves the Pine Editor to the right-side panel when required by the TradingView UI
- attempts to restore the editor to the bottom afterward
- verifies the actual editor placement instead of treating a menu-item click as successful
- returns `restored_to_bottom: false` if TradingView ignores the move operation

### Reliable Pine Editor opening

`ensurePineEditorOpen()` now uses a staged opening sequence:

1. checks whether a usable Pine Monaco editor is already available
2. tries TradingView's internal bottom-widget API
3. waits briefly for Monaco to mount before taking another action
4. checks whether the Pine side panel is already visible to avoid toggling it closed
5. resolves the visible Pine toolbar button's center coordinates
6. dispatches real CDP `mouseMoved`, `mousePressed`, and `mouseReleased` events
7. polls until Monaco is accessible

The previous synthetic `HTMLElement.click()` fallback has been removed. This matches the real input path that TradingView Desktop accepts in the affected panel state.

### Safer Monaco source updates

`pine_set_source` now updates the Monaco model through `executeEdits()` instead of `setValue()`.

The implementation now:

- selects the connected and visible Pine Monaco editor
- prefers the editor that currently has text focus
- normalizes line endings to the model's configured EOL format
- uses a full-model edit so TradingView receives an editor change event
- verifies the resulting model source
- reports whether the operation actually changed the source

This makes TradingView more likely to correctly mark the script as modified and enable the Save command.

### Reliable Pine Editor menu selection

The Pine context-menu trigger is now resolved from the Pine Editor tab's own DOM container.

It no longer searches all context-menu buttons and selects one based on Y-coordinate proximity. This avoids opening the Strategy Tester menu when both tabs are aligned at the same vertical position.

### First-time save support

`pine_save` now accepts an optional script name:

```json
{
  "name": "My Strategy"
}
```

The name is required when saving a new script for the first time.

When TradingView displays the first-save dialog, the implementation now:

1. locates the visible `Save script` dialog
2. finds its script-name input
3. updates the input using the native `HTMLInputElement.value` setter
4. dispatches `input` and `change` events
5. verifies that the Save button is enabled
6. submits the dialog

The CLI supports the same workflow:

```bash
tv pine save "My Strategy"
```

### Existing-script save protection

For an existing saved script, `pine_save` checks the currently attached script identity before saving.

If a different name is supplied, the operation is rejected instead of silently renaming, copying, or overwriting another script. `pine_save` is intentionally not treated as a rename or “save as” operation.

### Server-side persistence verification

After saving, the implementation reads the saved script through TradingView's Pine facade and compares its normalized source with the current editor source.

A save is only reported as successful when the server-side source matches the editor.

Possible successful actions include:

- `already_persisted`
- `saved_via_script_menu`
- `created_and_saved_via_dialog`

If the Save command is disabled, the operation only returns `already_persisted` when the saved server source already matches the editor. If the sources differ, the operation returns an error instead of reporting false success.

### New `pine_get_saved_source` tool

This PR adds `pine_get_saved_source`, which reads a saved Pine Script from TradingView without modifying the current editor.

It returns information including:

- saved script name and title
- script ID and version
- modification time
- source and line count

This provides a non-destructive way to verify persistence after `pine_save`.

### Clarified `pine_open` behavior

`pine_open` does not switch the TradingView script identity. It downloads a saved source and injects it into the current editor, which is a destructive operation.

Its description and return value now explicitly warn that it:

- overwrites the current editor source
- does not attach the editor to the selected saved script
- must not be used to verify that a save succeeded

## Screenshot tools disabled

This PR disables screenshot and image-recognition operations from the MCP server's structured tool surface.

### `capture_screenshot`

`registerCaptureTools()` is no longer registered by the MCP server. As a result, `capture_screenshot` is no longer exposed as an available MCP tool.

The capture implementation file is not deleted, but the server no longer advertises or registers the tool.

### Batch screenshot action

The screenshot action has also been removed from `batch_run`.

Previously accepted actions included:

```text
screenshot
get_ohlcv
get_strategy_results
```

The allowed actions are now restricted to:

```text
get_ohlcv
get_strategy_results
```

The tool schema now uses an enum so unsupported action names are rejected during argument validation. The batch screenshot implementation and its filesystem-writing dependencies were removed from `src/core/batch.js`.

### Server guidance

The MCP server instructions now explicitly state that:

- operations should use structured TradingView tools
- screenshots and image recognition are disabled
- a structured-tool failure should be reported directly instead of falling back to visual inspection

This prevents agents from claiming that a screenshot fallback was performed when no supported structured screenshot interface is available.

### `pine_save`

Before:

```text
pine_save()
```

Now:

```text
pine_save({ name?: string })
```

The name is required for the first save and optional when the current saved script identity can be determined.

### CLI save command

Both forms are now supported:

```bash
tv pine save
tv pine save "My New Script"
```

### `pine_new`

The tool description now clarifies that it creates a genuine TradingView script identity and refuses to discard unsaved changes.

### `batch_run`

The action argument is now validated as:

```text
get_ohlcv | get_strategy_results
```

The previous `screenshot` action is no longer available.

## Safety behavior

The updated workflow avoids the following unsafe outcomes:

- overwriting a saved script while pretending to create a new one
- saving through the Strategy Tester context menu
- reporting save success before server persistence
- accepting a first-save name without entering it into the dialog
- silently using a supplied name to rename or copy an existing script
- treating a Monaco source replacement as a registered TradingView edit
- using `pine_open` as proof that a script was saved
- reporting that the editor was restored to the bottom when only the menu item was clicked

## Tests

The regression coverage checks that:

- a closed Pine panel uses real CDP mouse input instead of `HTMLElement.click()`
- an already visible Pine panel is not toggled closed
- blank dirty buffers can be replaced while non-empty unsaved source remains protected
- Monaco updates use `executeEdits()` instead of `setValue()`
- line endings are normalized using the Monaco model EOL
- the resulting editor source is verified
- disabled Save commands do not report false success
- the Pine Editor menu is resolved from its own tab container
- the first-save dialog is filled and submitted
- `pine_new` uses the real `Create new` workflow
- `pine_new` no longer creates a template through `setSource()`
- editor placement restoration is verified
- saved-source verification is non-destructive
- `pine_open` is documented as destructive
- screenshot tools are not registered
- `batch_run` no longer accepts `screenshot`

Test results:

```text
npm run test:unit
163 tests passed
0 tests failed
```

Additional checks:

```text
node --check src/core/pine.js
git diff --check
npm run lint
```

ESLint reports 0 errors. Four existing unrelated warnings remain in `src/core/data.js`, `src/core/pane.js`, and `src/tools/watchlist.js`.

## Live TradingView verification

The updated local implementation was connected to a live TradingView Desktop session.

### Closed Pine Editor recovery

The Pine side panel was closed through its visible `Close` control and verified as hidden. Calling the updated local `ensurePineEditorOpen()` then reopened it through CDP mouse input and made Monaco accessible in approximately 6.1 seconds.

The follow-up source read returned the current blank buffer successfully:

```json
{
  "success": true,
  "source": "",
  "line_count": 1,
  "char_count": 0
}
```

### Blank and non-empty dirty-buffer behavior

From the blank dirty `Untitled script`, `pine_new({ type: "indicator" })` successfully created a genuine unsaved Indicator identity. No cloud save was performed.

A subsequent attempt to create a Strategy while the generated Indicator source was non-empty was rejected with the existing unsaved-changes error, and the source remained unchanged.

### Existing saved script

Saving the existing script returned:

```json
{
  "success": true,
  "action": "already_persisted",
  "name": "五线 EMA（可配置）",
  "script_id": "USER;e4e6a0fc7089424d956b556fea157f0f",
  "version": "4.0",
  "verified": true
}
```

### New Strategy identity

Creating a Strategy returned:

```json
{
  "success": true,
  "type": "strategy",
  "action": "new_tradingview_script_created",
  "identity_created": true,
  "previous_name": "五线 EMA（可配置）",
  "name": "Untitled script",
  "changed": true
}
```

The TradingView editor switched from the existing saved script to a genuine new `Untitled script` Strategy identity.

The temporary Strategy was not saved to the TradingView account, avoiding the creation of test scripts in the user's cloud script list.

## Compatibility notes

This PR intentionally changes the public MCP surface:

- `capture_screenshot` is no longer registered
- `batch_run(action: "screenshot")` is no longer supported
- first-time `pine_save` calls must provide a script name
- `pine_new` rejects non-empty unsaved content but allows recovery from a blank dirty buffer
- `pine_open` remains available but is explicitly documented as destructive source injection

### DIFF
--- a/src/cli/commands/pine.js
+++ b/src/cli/commands/pine.js
@@ -73,8 +73,8 @@ register('pine', {
       },
     }],
     ['save', {
-      description: 'Save the current Pine Script (Ctrl+S)',
-      handler: () => core.save(),
+      description: 'Save the current Pine Script; pass a name for the first save',
+      handler: (opts, positionals) => core.save({ name: positionals.join(' ') || undefined }),
     }],
     ['new', {
       description: 'Create a new blank Pine Script (indicator, strategy, library)',
--- a/src/core/batch.js
+++ b/src/core/batch.js
@@ -1,14 +1,8 @@
 /**
  * Core batch execution logic.
  */
-import { evaluate, evaluateAsync, getClient, getChartApi, getChartCollection, safeString } from '../connection.js';
+import { evaluate, evaluateAsync, getChartApi, getChartCollection, safeString } from '../connection.js';
 import { waitForChartReady } from '../wait.js';
-import { writeFileSync, mkdirSync } from 'fs';
-import { join, dirname } from 'path';
-import { fileURLToPath } from 'url';
-
-const __dirname = dirname(fileURLToPath(import.meta.url));
-const SCREENSHOT_DIR = join(dirname(dirname(__dirname)), 'screenshots');
 
 export async function batchRun({ symbols, timeframes, action, delay_ms, ohlcv_count }) {
   const tfs = timeframes && timeframes.length > 0 ? timeframes : [null];
@@ -35,16 +29,7 @@ export async function batchRun({ symbols, timeframes, action, delay_ms, ohlcv_co
         await new Promise(r => setTimeout(r, delay));
 
         let actionResult;
-        if (action === 'screenshot') {
-          mkdirSync(SCREENSHOT_DIR, { recursive: true });
-          const client = await getClient();
-          const { data } = await client.Page.captureScreenshot({ format: 'png' });
-          const ts = new Date().toISOString().replace(/[:.]/g, '-');
-          const fname = `batch_${symbol}_${tf || 'default'}_${ts}`.replace(/[\/\\]/g, '_') + '.png';
-          const filePath = join(SCREENSHOT_DIR, fname);
-          writeFileSync(filePath, Buffer.from(data, 'base64'));
-          actionResult = { file_path: filePath };
-        } else if (action === 'get_ohlcv' && apiPath) {
+        if (action === 'get_ohlcv' && apiPath) {
           const limit = Math.min(ohlcv_count || 100, 500);
           actionResult = await evaluateAsync(`
             new Promise(function(resolve, reject) {
--- a/src/core/pine.js
+++ b/src/core/pine.js
@@ -60,14 +60,20 @@ const FIND_MONACO = `
  * Returns true if editor is accessible, false on timeout.
  */
 export async function ensurePineEditorOpen() {
-  const already = await evaluate(`
-    (function() {
-      var m = ${FIND_MONACO};
-      return m !== null;
-    })()
-  `);
-  if (already) return true;
+  async function waitForEditor(attempts, delay = 100) {
+    for (let i = 0; i < attempts; i++) {
+      const ready = await evaluate(`(function() { var m = ${FIND_MONACO}; return m !== null; })()`);
+      if (ready) return true;
+      await new Promise(r => setTimeout(r, delay));
+    }
+    return false;
+  }
 
+  if (await waitForEditor(1, 0)) return true;
+
+  // Prefer TradingView's internal panel API, but give the editor time to mount
+  // before falling back to UI input. Some Desktop builds expose these methods
+  // while silently ignoring them for the right-side Pine panel.
   await evaluate(`
     (function() {
       var bwb = window.TradingView && window.TradingView.bottomWidgetBar;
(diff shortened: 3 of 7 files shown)


huof6829:
pine_save：
截屏2026-08-01 10 55 22

huof6829:
Screenshot：
截屏2026-07-31 23 36 48

PR #425, opened 2026-08-04
Fix MSIX detection when powershell.exe is absent from PATH

VGSgithub:
MSIX install detection shells out to `powershell -NoProfile -Command "(Get-AppxPackage ...)"`. On machines that ship PowerShell 7 without powershell.exe on PATH, that call fails outright. The error is discarded by the surrounding catch (and by 2^>nul in the .bat), so detection silently falls through and tv_launch reports "TradingView not found" even though the package is installed and resolvable.

Try `powershell` first, since it runs Get-AppxPackage natively and is fastest, then fall back to `pwsh`. Windows PowerShell stays the primary path, so behaviour is unchanged wherever it is present.

Also allow more headroom on the retry: pwsh reaches Get-AppxPackage through the WinPS compatibility layer, which is slower to start than 5.1.

Adds three regression tests covering host ordering, the pwsh fallback, and the neither-host-available error path.


VGSgithub (commented):
Stumbled upon this with claude code ... this may have been addressed already...

PR #429, opened 2026-08-04
feat: harden dangerous capabilities behind explicit opt-in gates

angeles725:
## Summary

Hardens the two dangerous MCP capabilities so they are **fail-closed by default** and only run after a deliberate, exact opt-in.

- `ui_evaluate` (arbitrary page JavaScript) and `replay_trade` (simulated Bar Replay position changes) are denied **before any CDP call** unless the server or CLI is started with an exact acknowledgment env var.
- Near-match and generic truthy values (`1`, `true`, lowercased ack, trailing space) are rejected — only the exact ack string enables the capability.
- The capability gate runs **before** schema validation, replay-API resolution, and CDP access, at the real MCP (`McpServer`) and CLI registration boundaries — not just at the core function.
- Replay navigation (`replay_start`, `replay_step`, `replay_autoplay`, `replay_status`, `replay_stop`) stays available without enabling simulated trades.

## Opt-in values

```text
TRADINGVIEW_MCP_ALLOW_ARBITRARY_PAGE_JS=I_UNDERSTAND_THIS_EXECUTES_ARBITRARY_JAVASCRIPT
TRADINGVIEW_MCP_ALLOW_REPLAY_TRADES=I_UNDERSTAND_THIS_CHANGES_SIMULATED_REPLAY_POSITIONS
```

## Docs

Corrected README/SECURITY claims: the MCP-to-Desktop transport is local, but the authenticated TradingView session and several tools can still reach TradingView services. The gate cannot prove demo/paper isolation — real brokers must be disconnected.

## Tests

- New `tests/capabilities.test.js` and expanded `tests/replay.test.js` cover default denial, near-match rejection, exact opt-in, valid-action pass-through, and both the SDK and registered-CLI boundaries.
- New `test:safe` npm script.
- Verified: `git diff --check` clean · 128 safe tests pass · lint 0 errors.

https://claude.ai/code/session_01UpZNatyXedsPXBXNEHJxGi

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -5,3 +5,10 @@ temp_*
 *.log
 discovery-log.json
 rules.json
+.atl/
+.claude/
+
+# Research-SDD investigation artifacts (not product)
+/corpus/
+/retros/
+/tools/
--- a/README.md
+++ b/README.md
@@ -77,6 +77,7 @@ Gives your AI assistant eyes and hands on your own chart:
 | `ui_evaluate` | Disabled | Exact arbitrary-JavaScript acknowledgment required |
 | `replay_trade` | Disabled | Exact simulated-position acknowledgment required |
 | `tv_update` | Disabled | Exact self-update acknowledgment required; pulls and runs remote code |
+| `alert_delete` (`delete_all`) | Enabled | Irreversible bulk delete requires an exact `confirm` token; a bare `delete_all` is refused |
 | Replay navigation | Enabled | `start`, `step`, `autoplay`, `status`, and `stop` do not require the trade capability |
 
 The exact opt-in values are documented in [SECURITY.md](SECURITY.md); generic truthy flags are rejected. `replay_trade` only changes TradingView's internal Bar Replay simulated positions. Replay and paper trading are not the same as enforced isolation: the MCP cannot verify account type or guarantee that a real broker is disconnected. Use replay only with real broker connections removed.
--- a/README.md
+++ b/README.md
@@ -76,6 +76,7 @@ Gives your AI assistant eyes and hands on your own chart:
 |------------|---------|----------|
 | `ui_evaluate` | Disabled | Exact arbitrary-JavaScript acknowledgment required |
 | `replay_trade` | Disabled | Exact simulated-position acknowledgment required |
+| `tv_update` | Disabled | Exact self-update acknowledgment required; pulls and runs remote code |
 | Replay navigation | Enabled | `start`, `step`, `autoplay`, `status`, and `stop` do not require the trade capability |
 
 The exact opt-in values are documented in [SECURITY.md](SECURITY.md); generic truthy flags are rejected. `replay_trade` only changes TradingView's internal Bar Replay simulated positions. Replay and paper trading are not the same as enforced isolation: the MCP cannot verify account type or guarantee that a real broker is disconnected. Use replay only with real broker connections removed.
--- a/README.md
+++ b/README.md
@@ -11,24 +11,24 @@ Personal AI assistant for your TradingView Desktop charts. Connects Claude Code 
 > **Requires a valid TradingView subscription.** This tool does not bypass or circumvent any TradingView paywall or access control. It reads from and controls the TradingView Desktop app already running on your machine.
 
 > [!NOTE]
-> **All data processing occurs locally on your machine.** No TradingView data is transmitted, stored, or redistributed externally by this tool.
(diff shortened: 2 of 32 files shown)


angeles725:
## Review — capability hardening & input robustness

Reviewed the full branch (6 commits) against `main`. **Verdict: approved, no blocking issues.**

The PR fail-closes the dangerous surfaces, guards one irreversible bulk operation, and validates several inputs that previously reached chart/network side effects unchecked.

### Protections added
| Surface | Mechanism | Default |
|---|---|---|
| `ui_evaluate` (arbitrary page JS) | Env-var capability gate + exact ack | Disabled |
| `replay_trade` (simulated positions) | Env-var capability gate + exact ack | Disabled |
| `tv_update` (pulls + runs remote code) | Env-var capability gate + exact ack | Disabled |
| `alert_delete` `delete_all` (irreversible bulk) | Per-call `confirm` token | Enabled, bulk needs token |
| `pine_get_source` (200KB+ payload) | Optional `max_chars` cap | Full source (unchanged) |
| `batch_run` / `pane` / `watchlist` | Input validation before side effects | — |

### Verified
- **Gate/guard runs before any side effect** in every path (core, MCP tool, CLI):
  - `ui_evaluate` / `replay_trade`: gate before any CDP call. Confirmed against the CLI router — handlers receive only `(values, positionals)`, so `_deps` is `undefined` in production and the gate falls back to `process.env`.
  - `tv_update`: gate is the first statement of `update()`, before any git/network/npm; a test asserts zero commands run when disabled.
  - `alert_delete`: `delete_all` refuses before the network list/delete requests unless `confirm: "DELETE_ALL_ALERTS"` is passed.
  - `batch_run` / `pane` / `watchlist`: inputs validated before switching a symbol, focusing a pane, or driving the watchlist UI/REST.
- **Exact-match acks**; near-match values (`true`, lowercased, trailing space) rejected for all three env gates and the confirm token.
- **No callers broken** — each gated function has only the MCP-tool and CLI call sites; no internal callers.
- **A real bug was caught and fixed during this work**: `Number(null) === 0` and `Number('') === 0` meant a nullish pane index silently targeted pane 0 and proceeded to live I/O; `requirePaneIndex` now rejects nullish/blank explicitly.
- **Tests**: 164 offline tests pass (`test:safe`), lint 0 errors, `git diff --check` clean. New suites: `capabilities`, `alerts`, `pine_source`, `input_validation`, plus expanded `replay` and `update`. Coverage includes the real SDK boundary (`InMemoryTransport`), registered CLI handlers, near-match rejection, and denial-before-side-effect.
- **Docs**: README Security Boundaries table + SECURITY.md document all three env gates, the confirm token, and the accurate data-locality boundary (the earlier "all data local / safe to run" claims were corrected).
- A read-only audit of the remaining tool surface found **no injection vulnerabilities**: every user-controlled parameter interpolated into a page-evaluated string is escaped (`JSON.stringify` / numeric coercion).

### Opt-in env vars
```text
TRADINGVIEW_MCP_ALLOW_ARBITRARY_PAGE_JS=I_UNDERSTAND_THIS_EXECUTES_ARBITRARY_JAVASCRIPT
TRADINGVIEW_MCP_ALLOW_REPLAY_TRADES=I_UNDERSTAND_THIS_CHANGES_SIMULATED_REPLAY_POSITIONS
TRADINGVIEW_MCP_ALLOW_SELF_UPDATE=I_UNDERSTAND_THIS_PULLS_AND_RUNS_REMOTE_CODE
```

### Non-blocking observations (by design)
1. Hardening is scoped to the surfaces above; other state-changing tools remain open and are now documented as such in SECURITY.md ("treat MCP clients and prompts as trusted code").
2. Env opt-in is process-global (whole server lifetime), no per-request consent — appropriate for a local single-user MCP.
3. `replay_trade` uses `z.unknown().optional()` so the gate error wins over Zod validation — a deliberate tradeoff that drops the enum schema hint.
4. Consistency follow-up (not in this PR): several core modules still lack the `_deps` injection seam; a future pass can standardize it for testability.

### Commits
1. `feat: harden dangerous capabilities behind explicit opt-in gates` — ui_evaluate + replay_trade
2. `fix: gate ui_evaluate before arg validation at the CLI boundary`
3. `feat: gate tv_update behind explicit self-update opt-in`
4. `feat: require explicit confirmation for alert_delete delete_all`
5. `feat: add optional output cap to pine_get_source`
6. `fix: validate inputs before side effects in batch_run, pane, watchlist`

PR #437, opened 2026-08-07
Raise the MSIX detection timeout so Store installs are actually found

maxvama-beep:
tv_launch fails with "TradingView not found on win32" on machines where TradingView is installed from the Microsoft Store. The error lists only the three standard filesystem paths, so the MSIX branch looks like it never ran.

It runs, and gets killed. Get-AppxPackage plus PowerShell startup measured 6.1s, 6.1s and 7.8s across three consecutive runs on an idle Windows 11 box (TradingView.Desktop 3.3.0.7992). The call is given a 5s budget, so execSync kills it with ETIMEDOUT every time. The catch { /* ignore */ } swallows it, detection falls through to where TradingView.exe — which doesn't find MSIX apps either — and the user gets a not-found error naming paths that were never relevant.

Raises the budget to 20s and adds -NonInteractive so the call can't block on a prompt and burn it. The slower failure only affects the path where the three standard locations already missed.

Verified end-to-end: detection resolves C:\Program Files\WindowsApps\TradingView.Desktop_3.3.0.7992_x64__n534cwy3pjxzj, launch succeeds and CDP binds in 14.4s total, no local-copy fallback needed.

ESLint clean. test:unit 123/125 — the two failures (compiles valid Pine Script, returns errors for invalid Pine Script, plus source audit) reproduce identically on clean main at c05b8f5 without this commit, so they're pre-existing and unrelated.

### DIFF
--- a/src/core/health.js
+++ b/src/core/health.js
@@ -318,8 +318,12 @@ export async function launch({ port, kill_existing, _deps } = {}) {
     // MSIX/Windows Store install — InstallLocation is in WindowsApps, which is ACL-restricted
     // for normal `dir` enumeration but readable via Get-AppxPackage without elevation.
     try {
-      const ps = 'powershell -NoProfile -Command "(Get-AppxPackage -Name \'TradingView.Desktop\' -ErrorAction SilentlyContinue).InstallLocation"';
-      const installDir = deps.execSync(ps, { timeout: 5000 }).toString().trim();
+      // 20s, not 5s: Get-AppxPackage plus PowerShell startup measured 6.1-7.8s
+      // on a normal Windows 11 box, so a 5s budget killed this call every time
+      // and MSIX installs were never found. -NonInteractive so it can never
+      // block on a prompt and burn the whole budget.
+      const ps = 'powershell -NoProfile -NonInteractive -Command "(Get-AppxPackage -Name \'TradingView.Desktop\' -ErrorAction SilentlyContinue).InstallLocation"';
+      const installDir = deps.execSync(ps, { timeout: 20000 }).toString().trim();
       if (installDir) {
         const candidate = `${installDir}\\TradingView.exe`;
         if (deps.existsSync(candidate)) tvPath = candidate;


L3G1T50:
Independent confirmation on a different Windows 11 machine, same package version (TradingView.Desktop 3.3.0.7992). I hit this before finding your PR — `tv launch` failed with the same not-found error naming the three non-MSIX paths, while `Get-AppxPackage -Name 'TradingView.Desktop'` resolved the install fine when run directly.

Three consecutive runs here:

```
run 1: 12162 ms
run 2:  8073 ms
run 3:  5440 ms
```

Same shape as your 6.1 / 6.1 / 7.8s, but with a noticeably slower cold run. Worth flagging: **12.2s against a 20s budget is less headroom than it looks**, because the slow run is the first one after boot — which is exactly when someone reaches for `tv_launch`. I landed on 30s locally for that reason. It only bounds a probe that returns in well under a second once the appx stack is warm, and it's only reached after the three standard paths have already missed, so the cost of a larger budget is nil on every other path.

The `-NonInteractive` addition is a good catch — I didn't have that.

One thing that might help this stick: a regression test pinning the budget so it can't silently drift back to 5s. This works against the existing `msixDeps` mock, so it needs no new fixtures, and it skips off-Windows like the other MSIX suites:

```js
it('gives Get-AppxPackage a budget that survives a cold appx stack', async () => {
  const calls = [];
  const { deps } = msixDeps({ cdpBindsFor: ['WindowsApps'] });
  const passthrough = deps.execSync;
  deps.execSync = (cmd, opts) => { calls.push({ cmd, opts }); return passthrough(cmd, opts); };

  await launch({ _deps: deps });

  const probe = calls.find((c) => c.cmd.includes('Get-AppxPackage'));
  assert.ok(probe, 'expected launch() to probe Get-AppxPackage on win32');
  assert.ok(
    probe.opts && probe.opts.timeout >= 15000,
    `Get-AppxPackage needs a timeout that tolerates a slow appx stack, got ${probe.opts && probe.opts.timeout}ms`,
  );
});
```

I verified it fails at 5000 and passes at the raised value. Happy to open it as a follow-up PR or leave it for you to fold in here — whichever you'd rather.

Verification on my end: detection resolves `C:\Program Files\WindowsApps\TradingView.Desktop_3.3.0.7992_x64__n534cwy3pjxzj`, launch succeeds, CDP binds on 9222 (Chrome/140.0.7339.133, Electron 38.2.2), no local-copy fallback needed. `tv status` then returns live chart state.

On the test suite: I also see 123/125 on `test:unit`. The source-audit failure is a test-side path bug — it builds `C:\C:\Users\...\src\core\` and ENOENTs — and the two `pine check` cases reach TradingView's server-compile endpoint. Neither goes near this code path.

L3G1T50:
Opened that regression test as a PR against your branch rather than upstream, so it lands together with the fix instead of sitting red on `main` waiting for it:

https://github.com/maxvama-beep/tradingview-mcp/pull/10

Test only — +21/−0, `tests/launch.test.js`, nothing touched in your `health.js`. It asserts `>= 15000` rather than your exact 20000, so the value stays yours to tune. Verified both directions: passes on your branch, fails when the timeout is reverted to 5000.

Entirely optional. Merge it and this PR carries fix + test together; ignore it and yours is unaffected either way.

maxvama-beep:
Thanks — independent repro on a second box is exactly what this needed, and
the cold-run number is the useful part. You've talked me into 30s. My 20s
was fitted to my own warm-ish runs; 12.2s on a cold stack makes that 1.6×
headroom on a sample of two machines, and the failure mode is asymmetric —
too short breaks the feature outright, too long costs ten seconds on a path
that's only reached after three misses. Raising it.

Please fold the test in here rather than as a follow-up — it only passes
because of this change, and I'd rather not have the fix sitting on main
unguarded. Two small things when you do:

   - Can you raise the floor to >= 20000? At 15000 the test passes at a
   budget below the one we're agreeing on, and only ~1.2× the 12.2s you
   measured — which undercuts what the test name claims.
   - The snippet as pasted doesn't carry the win32 guard. If it's going
   inside the existing guarded describe that's fine, just flagging it so it
   doesn't trip the non-Windows CI.

Also checking whether anything upstream wraps this with its own budget — if
the tool call has a 30s ceiling of its own, a 30s inner probe would race it
and I'd rather find that now than in a bug report.

Agreed on the two suite failures: the source-audit one is a test-side path
join against an already-absolute path, and the pine check cases hitting the
live compile endpoint means test:unit isn't hermetic. Both worth their own
issues, neither near this path. I'll open them unless you've already got
them filed.


With Best Regards

*S.Uma Maheshwar Sastry*

L3G1T50:
I hit this same bug independently and had a fix sitting on a branch — then found this PR, which predates mine by a week and gets the diagnosis right. Not opening a competing one. `-NonInteractive` is a good catch that mine didn't have.

Offering the one piece mine has that this doesn't: a regression test. It's written to assert `>= 15000`, not an exact value, so **it passes against your 20000ms as-is** — it locks in the class of fix rather than my particular number.

```js
describe('launch() — MSIX detection budget', { skip: !onWindows }, () => {
  // Get-AppxPackage routinely takes 5-12s on a cold appx stack. While this budget
  // was 5000ms it timed out on real machines; the error was swallowed and every
  // MSIX install reported "TradingView not found" against three irrelevant paths.
  it('gives Get-AppxPackage a budget that survives a cold appx stack', async () => {
    const calls = [];
    const { deps } = msixDeps({ cdpBindsFor: ['WindowsApps'] });
    const passthrough = deps.execSync;
    deps.execSync = (cmd, opts) => { calls.push({ cmd, opts }); return passthrough(cmd, opts); };

    await launch({ _deps: deps });

    const probe = calls.find((c) => c.cmd.includes('Get-AppxPackage'));
    assert.ok(probe, 'expected launch() to probe Get-AppxPackage on win32');
    assert.ok(
      probe.opts && probe.opts.timeout >= 15000,
      `Get-AppxPackage needs a timeout that tolerates a slow appx stack, got ${probe.opts && probe.opts.timeout}ms`,
    );
  });
});
```

**Verified both directions** on Windows 11 / Node v24.14.1, by checking out `origin/main`, applying this PR's diff, and dropping the test in:

- with this PR's `timeout: 20000` → `node --test tests/launch.test.js` exits **0**, budget test green
- reverting only `src/core/health.js` to `main`'s `timeout: 5000` → exits **1**, with
  `AssertionError: Get-AppxPackage needs a timeout that tolerates a slow appx stack, got 5000ms`

So it genuinely guards the bug rather than just passing alongside the fix.

It needs **no new scaffolding** — it reuses the existing `msixDeps` helper already defined at `tests/launch.test.js:32` and drops in after the `MSIX WindowsApps handling` block.

One corroborating data point for the PR body: I measured `Get-AppxPackage` at **12.2s / 8.1s / 5.4s** across three consecutive cold runs (TradingView.Desktop 3.3.0.7992). That's a bit worse than the 6.1–7.8s cited here, though still well inside the 20s budget — just extra evidence that 5s was never survivable.

Happy either way: copy the block straight into this PR (no attribution needed), or I'll open a test-only PR on top once this merges. Whichever is less work for you.

L3G1T50:
Upstream CI never ran on this PR. Its only workflow run ([31155905311](https://github.com/tradesdontlie/tradingview-mcp/actions/runs/31155905311)) was never approved. It now shows `failure` with no jobs, set 30 days after the run was created. A new run would need a new push or a close/reopen.

I ran the same workflow in my fork against this PR's commit, `4e840c8`, whose parent is the current `main` (`c05b8f5`):

https://github.com/L3G1T50/tradingview-mcp/actions/runs/35158234941 (`success`)

The run used a throwaway branch: `4e840c8` plus one commit that only adds `'ci/**'` to the `push` trigger in `.github/workflows/ci.yml`, so the workflow would run on push. The jobs and steps are unchanged, and no other file differs from this PR.

- `npm ci`: passed.
- `npm run test:unit`: 152/152 passed.

This run doesn't test the timeout change itself. The `launch()` suites in `tests/launch.test.js` skip on anything but Windows, and both of them (`MSIX WindowsApps handling` and `classic install path`) were skipped here. So it only shows that the change doesn't break lint or the rest of the suite; only a Windows run covers the MSIX path.

L3G1T50:
Following up on the Ubuntu run above: the `launch()` suites skip off Windows, so I ran the suite on Windows too. The setup was Windows 11 (build 26200) and Node v24.19.0, with `npm ci` on this PR's lockfile, which is the same as `main`'s.

`npm run test:unit` on this PR (`4e840c8`) and on `main` (`c05b8f5`) gave identical results: 123/125 passed.
- The 2 failures are the `CLI — pine check (server compile)` tests. The CLI exits with code 3221226505 (`0xC0000409`) on both commits, so it's not related to this PR.
- `source audit — no unsafe interpolation patterns` fails before any of its tests run, because of the doubled `C:\C:\` path, and it does the same on `main`. That's why Windows counts 125 tests against Ubuntu's 152: the 34 source-audit tests never run, and 7 `launch()` tests run only here.
- All 7 `launch()` tests passed: 5 in `MSIX WindowsApps handling` and 2 in `classic install path`. This is the only run that exercises the changed code path.

Those suites don't check the timeout, though. The `msixDeps` mock answers any `Get-AppxPackage` command and ignores the options, so they pass at 5000 ms too. The regression test I offered above covers that; it's still open against your branch as maxvama-beep/tradingview-mcp#10. With it on top of this PR (`15c0135`, the head of that PR), `node --test tests/launch.test.js` passed 8/8. The same test file against `main`'s `health.js` failed as expected: `got 5000ms`.

PR #439, opened 2026-08-08
Find the Pine compile button across locales

Maikstyler79:
## Problem

`pine_smart_compile` / `pine_compile` locate the compile button by matching `button.textContent` against the English strings `"Add to chart"` / `"Update on chart"`.

Two things break that on a non-English TradingView UI:

1. **The labels are localised.** On a German UI the button reads `Dem Chart hinzufügen`.
2. **The desktop build renders the label twice inside the button**, so `textContent` is literally `"Dem Chart hinzufügenDem Chart hinzufügen"` — which also defeats the anchored `/^add to chart$/i` match even in English in some builds.

When nothing matches, the current code falls through to:

```js
if (saveBtn) { saveBtn.click(); return 'Pine Save'; }
```

So instead of compiling, it **clicks Save** — writing the script to the user's TradingView cloud — and the caller sees `success: true`. On a German desktop client every compile silently became a save.

## Fix

A single shared `FIND_COMPILE_BUTTON` helper used by both `compile()` and `smartCompile()`:

- matches against **`textContent`, `title` and `aria-label`** (the attributes carry the un-doubled label)
- **collapses the doubled label** (`"FooFoo"` → `"Foo"`) before comparing
- covers the common TradingView UI languages (de, en, fr, es, it, pt, ru, zh, ja)
- skips invisible (`offsetParent === null`) and disabled buttons, so it only ever hits the visible editor
- **prefers the plain add/update buttons.** `Save and add to chart` is used only when nothing else is present, and is reported under its own name — a cloud save is never reported as a plain compile
- **drops the plain Save button as a fallback entirely.** Saving is not compiling; reporting it as success hides a failed compile behind a write to the user's account

The keyboard-shortcut fallback (`Ctrl+Enter`) is unchanged and still applies when no button is found.

## Verification

Tested against **TradingView Desktop 3.3.0 (MSIX) on Windows 11 with a German UI**, driven over CDP:

- button resolution on the live DOM:
  `{"chosen":"Add to chart","addBtn":{"text":"Dem Chart hinzufügenDem Chart hinzufügen","title":"Dem Chart hinzufügen"}}`
- `pine_smart_compile` on a Pine v6 indicator with two EMAs →
  `{"button_clicked":"Add to chart","has_errors":false,"study_added":true}`
- indicator confirmed present via `chart_get_state` (`entity_id` returned), then removed again
- no save dialog opened at any point

`npm run lint` clean (0 errors). `tests/pine_analyze.test.js` + `tests/sanitization.test.js`: 50/50 pass.

## Note on Windows test runs (not part of this PR)

Two pre-existing, unrelated failures show up on Windows and are worth a separate issue:

- `tests/e2e.test.js` — the `tv_launch` case asserts macOS-only paths (`/Applications/TradingView.app/...`), so it can never pass on Windows or Linux.
- `tests/sanitization.test.js` — `new URL('../src/core/', import.meta.url).pathname` yields `/C:/...`, so `readdirSync` resolves to `C:\C:\Users\...` and the suite errors during construction. `fileURLToPath()` would fix it.

### DIFF
--- a/src/core/pine.js
+++ b/src/core/pine.js
@@ -5,6 +5,60 @@
  */
 import { evaluate, evaluateAsync, getClient } from '../connection.js';
 
+/**
+ * Compile-button finder (injected into TV page).
+ *
+ * TradingView localises the Pine editor buttons, and the desktop build renders
+ * the label twice inside the button ("Add to chartAdd to chart"), so matching
+ * on visible textContent alone fails outside en-US. Check title and aria-label
+ * too — those carry the un-doubled label — and de-duplicate the text before
+ * comparing. Returns the canonical English action so callers stay locale-free.
+ *
+ * Deliberately never falls back to the plain Save button: saving is not compiling.
+ * "Save and add to chart" is used only when no plain add/update button exists, and
+ * is reported under its own name so callers can tell that a cloud save happened.
+ */
+const FIND_COMPILE_BUTTON = `
+  (function findCompileButton() {
+    var ADD = /(add to chart|dem chart hinzuf|zum chart hinzuf|ajouter au graphique|agregar al gr|aggiungi al grafico|adicionar ao gr|добавить на график|添加到图表|グラフに追加)/i;
+    var UPDATE = /(update on chart|auf dem chart aktualisier|chart aktualisier|mettre . jour sur le graphique|actualizar en el gr|aggiorna sul grafico|atualizar no gr|обновить на графике|更新图表)/i;
+    var SAVE_AND_ADD = /(save and add to chart|speichern und dem chart hinzuf)/i;
+
+    function labelsOf(el) {
+      var txt = (el.textContent || '').trim().replace(/\\s+/g, ' ');
+      // Collapse the doubled label ("FooFoo" -> "Foo").
+      var half = txt.length / 2;
+      if (txt.length > 1 && txt.length % 2 === 0 && txt.slice(0, half) === txt.slice(half)) {
+        txt = txt.slice(0, half);
+      }
+      return [txt, el.getAttribute('title') || '', el.getAttribute('aria-label') || ''];
+    }
+
+    var btns = document.querySelectorAll('button,[role="button"]');
+    var addBtn = null, updateBtn = null, saveAddBtn = null;
+
+    for (var i = 0; i < btns.length; i++) {
+      var el = btns[i];
+      if (el.offsetParent === null) continue;      // only the visible editor
+      if (el.disabled) continue;
+      var ls = labelsOf(el);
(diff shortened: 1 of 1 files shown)


palmi85:
I hit this independently on TradingView Desktop 3.3.0 with a German UI (macOS 26.5, CDP on 127.0.0.1:9222) and can confirm the diagnosis exactly as written here, including the doubled `textContent`. Before any fix, `pine_smart_compile` returned:

```json
{"success": true, "button_clicked": "Pine Save", "has_errors": false, "errors": [], "study_added": false}
```

Worth adding: it doesn't just save silently — it leaves a modal `Script speichern` dialog open on the chart, so the *next* tool call runs against a blocked UI. For an unsaved script the save doesn't even complete; you get a name prompt sitting there.

One suggestion for the selector. Rather than maintaining a language list, that button carries a stable test hook:

```
data-qa-id="add-script-to-chart"
```

I dumped every `[data-qa-id*=script]` in the editor toolbar on the German build:

| `data-qa-id` | title |
|---|---|
| `pine-script-title-button` | — |
| `add-script-to-chart` | `Dem Chart hinzufügen` |
| `pine-script-save-button` | `Skript speichern` |
| `publish-script` | `Skript veröffentlichen` |
| `script-more-options` | `Mehr` |

Two things that make this convenient:

- The id is **identical in the add and update states** — TradingView only swaps the `title` to `Auf dem Chart aktualisieren` once the script is on the chart. So a single selector covers both, no separate update branch needed.
- In the update state the button is icon-only (`textContent` is `""`), so any text-based match has to fall back to `title`/`aria-label` there regardless.

It also sidesteps the near-miss risk of matching `publish-script`, whose label (`Skript veröffentlichen`) sits right next to the compile button in the DOM.

I'd keep the localized-text matching from this PR as the fallback for older builds, just behind the `data-qa-id` lookup. And the most important part of this PR, independent of how the button is found, is deleting the `saveBtn` fallback — a save dialog is not a compile, and silently returning `success: true` for it is what makes the bug so hard to spot.

Verification on my side, with the `data-qa-id` lookup in front of the text match:

- add: `button_clicked: "Add to chart"`, `study_added: true`, study visible in `chart_get_state`
- update: `button_clicked: "Update on chart"` (label `Auf dem Chart aktualisieren`), no save dialog
- `npm run lint` clean; `npm run test:all` 230/231, the one failure (`replay_start`) is unrelated and identical on `main` — intraday Bar Replay is gated behind a paid plan on my account, so TradingView shows an upsell modal instead of starting replay.

Happy to push my version as a commit here if that's useful, or just leave the selector suggestion for you to fold in — your call, no interest in a competing PR.

PR #445, opened 2026-08-09
fix: resolve npm audit vulnerabilities via dependency updates

AlxndrJW:
Bumps transitive deps (hono, body-parser, brace-expansion, fast-uri, ip-address, js-yaml) to patched versions, clearing all 8 reported vulnerabilities (4 high, 3 moderate, 1 low).

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -187,12 +187,12 @@
       }
     },
     "node_modules/@hono/node-server": {
-      "version": "1.19.14",
-      "resolved": "https://registry.npmjs.org/@hono/node-server/-/node-server-1.19.14.tgz",
-      "integrity": "sha512-GwtvgtXxnWsucXvbQXkRgqksiH2Qed37H9xHZocE5sA3N8O8O8/8FA3uclQXxXVzc9XBZuEOMK7+r02FmSpHtw==",
+      "version": "2.1.0",
+      "resolved": "https://registry.npmjs.org/@hono/node-server/-/node-server-2.1.0.tgz",
+      "integrity": "sha512-XovyyCCnBzW+zKu+z/zq8hwNs4KOR5rEMAOxo2f40Q5xoOI37IMm6MIg2COOUtUApo0i6850MTBKH2u4QLGIqg==",
       "license": "MIT",
       "engines": {
-        "node": ">=18.14.1"
+        "node": ">=20"
       },
       "peerDependencies": {
         "hono": "^4"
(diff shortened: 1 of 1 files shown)


L3G1T50:
Independent corroboration — I hit the same advisories and arrived at an almost identical lockfile before finding this PR, so this is a confirmation rather than a competing change.

I fetched this PR's head and audited it directly:

```
git fetch origin refs/pull/445/head
npm audit --package-lock-only
→ found 0 vulnerabilities
```

So this still clears everything as of today, not just when it was opened.

Comparing your lockfile against the one `npm install` produced for me independently: same 187-line diff, same bumps, with exactly one difference —

| | `c05b8f5` (base) | this PR | mine |
| --- | --- | --- | --- |
| `ip-address` | 10.2.0 | 10.4.0 | 10.5.0 |

Everything else matches, `espree` included (10.4.0 on both — unchanged from base). `ip-address` is one of the two highs, and since your 10.4.0 already audits clean, the extra patch bump on my side is immaterial. No reason to prefer mine.

One note on reproducibility: I measure **7** advisories on the base commit (4 high / 2 moderate / 1 low) where the description says 8 (4 / 3 / 1). Most likely advisory-database drift between Aug 9 and now rather than a difference in the tree — the high-severity set is identical either way (`brace-expansion`, `fast-uri`, `ip-address`, `js-yaml`).

Worth flagging separately, since it explains how a lockfile with 4 highs went unnoticed: the audit step in `.github/workflows/ci.yml` is set to `continue-on-error: true`, so `npm audit --audit-level=high` can never fail the build. Dropping that flag — or at minimum failing on `critical` — would catch the next drift automatically instead of relying on someone noticing. Happy to open that as its own PR if it'd be useful; it's independent of this change and shouldn't be bundled in here.

Also relevant: #389 covers part of the same ground but is superseded. It fixed 4 of 6 and explicitly left the `@hono/node-server` moderates as unfixable ("even the latest SDK release still pins it to a pre-patch major version"). That's no longer true — `@modelcontextprotocol/sdk` 1.30.0 pulls `@hono/node-server` 2.1.0, which is what this PR does and why it gets to zero. It's also bundled with an unrelated `.gitignore` change. This PR looks like the one to take.

Environment: Windows 11, Node 24.14.1, npm audit against `package.json` unchanged (`^1.12.1` already permitted every one of these resolutions — no manifest edit needed).

L3G1T50:
One consequence of this PR that isn't visible in the diff — flagging it for the record, not asking you to change anything.

The `@hono/node-server` bump carries an engines change:

```
base   @hono/node-server=1.19.14  engines={"node":">=18.14.1"}
this   @hono/node-server=2.1.0    engines={"node":">=20"}
```

So merging this raises the project's supported Node floor to 20. `README.md:53` still advertises **`Node.js 18+`**, and that line is the only place the floor is stated anywhere — `package.json` has no `engines` field — so it goes stale on merge.

**Nothing to do here.** This PR is `package-lock.json` only and that's the correct scope; a docs change doesn't belong in it. I've sent the one-line README fix as #478, written so it carries **no ordering dependency** on this PR: it's already correct on `main` as it stands (CI has only ever tested `[20.x, 22.x]`, so Node 18 was advertised but never exercised), and it stays correct after this merges. Either order, or neither — this PR isn't blocked by it.

Also closing a loop from my earlier comment: I offered to send the `continue-on-error: true` removal separately, and did — that's #454. It does depend on a lockfile refresh landing first, and its body names this PR as the prerequisite, since the gate can't go green until the advisories clear.

PR #457, opened 2026-08-13
chore(deps): refresh lockfile to clear 7 advisories, raise Node floor to 20+

L3G1T50:
`npm audit` reports **7 vulnerabilities against the committed lockfile — 4 of them high**, and 5 of the 7 (including 2 high) reach production dependencies. Every one is transitive, reached through `@modelcontextprotocol/sdk`, and every one clears by re-resolving inside the range `package.json` already declares.

**`package.json` is unchanged.** The declared `^1.12.1` range already permitted these versions; the lockfile was simply pinned to older resolutions.

## Before

```
$ npm audit --package-lock-only
7 vulnerabilities (1 low, 2 moderate, 4 high)
```

| severity | package | reaches |
| --- | --- | --- |
| high | `fast-uri` | **production** |
| high | `ip-address` | **production** |
| high | `brace-expansion` | dev only |
| high | `js-yaml` | dev only |
| moderate | `hono` | **production** |
| moderate | `@hono/node-server` | **production** |
| low | `body-parser` | **production** |

Production-only run isolates the reachable subset:

```
$ npm audit --package-lock-only --omit=dev
5 vulnerabilities (1 low, 2 moderate, 2 high)
```

## After

```
$ npm audit --package-lock-only
found 0 vulnerabilities
```

Three resolutions move:

```
@modelcontextprotocol/sdk  1.27.1  -> 1.30.0
@hono/node-server          1.19.14 -> 2.1.0
hono                       4.12.27 -> 4.13.1
```

## This raises the supported Node floor — please read

`@hono/node-server@2.1.0` declares `engines: { node: ">=20" }`, up from `>=18`.

- CI is unaffected — the matrix is already `[20.x, 22.x]`.
- But `README.md:53` advertised **Node.js 18+**, which the lockfile change quietly invalidates. `package.json` has no `engines` field, so nothing enforces or contradicts the floor in code; that README line was the only statement of it.

**So this PR bumps the README to Node 20+ in a second commit** (`docs: raise the documented Node floor to 20+`), kept separate from the lockfile diff so the user-facing change is visible in its own right rather than buried in a resolution churn.

The alternative, if dropping Node 18 is unacceptable: **hold `@hono/node-server` at 1.x**, keep the 18+ claim, and accept the two moderate `hono`-chain advisories. The two *high* production ones (`fast-uri`, `ip-address`) come through the SDK bump and would still clear either way. Say the word and I'll rework it that way.

I went with the bump because Node 18 reached end-of-life in April 2025 and CI has not tested it for some time — so the 18+ claim was already aspirational rather than verified.

## Verification

Run on Windows 11, Node 24.14.1:

- `npm ci` clean
- `eslint` 0 errors
- `npm run test:unit` 124/126 — the 2 failures are the pre-existing `pine check` cases that reach TradingView's server-compile endpoint, and they fail identically on `main`
- MCP server module loads against the new SDK

The before/after audit figures above were re-measured against `origin/main` today, not carried over from when the lockfile was regenerated.

## Relationship to #454

#454 makes the high-severity audit gate blocking in CI. On its own it would **fail** the build against the current lockfile, because of the 4 highs above. This PR is what makes that gate pass. They're independent changes but land best together — merging this one first is the natural order.

---

## CI verification

Upstream CI has never run on this PR. Like every external PR on this repo since 2026-07-24, its workflow run is held at `action_required` — 92 runs from 49 contributors are in the same state (details in #474). So the change is not left unverified, I ran the identical workflow on my fork:

**https://github.com/L3G1T50/tradingview-mcp/actions/runs/32521718142** — `success`, both matrix legs:

- `lint-and-test (20.x)` — success
- `lint-and-test (22.x)` — success

That run executed `.github/workflows/ci.yml` at blob `1a5e337`, which is **byte-identical to the copy on `tradesdontlie:main`** (I compared the blob SHAs, not just the contents) — same workflow, unmodified, ubuntu-latest, lint + unit tests + audit.

**Caveat, stated plainly — this one needs two qualifications:**

1. The lockfile change is in the tested tree as `159abc0`, a different SHA from this branch's `7acd221`. They are the same change: `git patch-id --stable` gives `8f6ab2db` for both. It was cherry-picked onto a different base, nothing more.
2. **The README commit (`6991e1e`, Node floor 18+ → 20+) is not in the tested tree.** It is documentation only and cannot affect the job, but the run does not cover it and I would rather say so than let the citation imply otherwise.

As with my other PRs, the tested tree also carries unrelated commits, so this is not per-PR isolation — supporting evidence, not a substitute for CI here.

### DIFF
--- a/README.md
+++ b/README.md
@@ -50,7 +50,7 @@ See [RESEARCH.md](RESEARCH.md) for open questions, findings, and related work.
 ## Prerequisites
 
 - **TradingView Desktop app** (paid subscription required for real-time data)
-- **Node.js 18+**
+- **Node.js 20+**
 - **Claude Code** with MCP support (for MCP tools) or any terminal (for CLI)
 - **macOS, Windows, or Linux**
 
--- a/package-lock.json
+++ b/package-lock.json
@@ -187,12 +187,12 @@
       }
     },
     "node_modules/@hono/node-server": {
-      "version": "1.19.14",
-      "resolved": "https://registry.npmjs.org/@hono/node-server/-/node-server-1.19.14.tgz",
-      "integrity": "sha512-GwtvgtXxnWsucXvbQXkRgqksiH2Qed37H9xHZocE5sA3N8O8O8/8FA3uclQXxXVzc9XBZuEOMK7+r02FmSpHtw==",
+      "version": "2.1.0",
+      "resolved": "https://registry.npmjs.org/@hono/node-server/-/node-server-2.1.0.tgz",
+      "integrity": "sha512-XovyyCCnBzW+zKu+z/zq8hwNs4KOR5rEMAOxo2f40Q5xoOI37IMm6MIg2COOUtUApo0i6850MTBKH2u4QLGIqg==",
       "license": "MIT",
       "engines": {
-        "node": ">=18.14.1"
+        "node": ">=20"
       },
       "peerDependencies": {
         "hono": "^4"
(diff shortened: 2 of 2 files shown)


L3G1T50:
Closing this in favor of **#445**, which does the same job and was filed first (2026-08-09; this one 2026-08-13). I didn't see it when I opened this — my mistake, and the duplicate work is on me.

For the record, I diffed the two lockfiles resolution by resolution rather than assuming they matched. Out of every entry in both files, they differ in **exactly one**:

| | `main` | #445 | this PR |
| --- | --- | --- | --- |
| `ip-address` | 10.2.0 | 10.4.0 | 10.5.0 |

Both clear all seven advisories:

```
npm audit --audit-level=high
  origin/main       -> exit 1  (7 vulnerabilities, 4 high)
  refs/pull/445/head -> exit 0  (found 0 vulnerabilities)
```

A patch-level difference in one transitive package is not a reason to prefer this branch, and #445 got there first. Nothing here is worth holding up.

## One thing in this PR that is not in #445, and still applies

#445 resolves `@hono/node-server` to `2.1.0`, same as this branch did:

```
main   @hono/node-server=1.19.14  engines={"node":">=18.14.1"}
#445   @hono/node-server=2.1.0    engines={"node":">=20"}
```

So **#445 raises the supported Node floor to 20**, and `README.md:53` still advertises **`Node.js 18+`**. #445 touches `package-lock.json` only, so that line goes stale the moment it merges. CI is unaffected (the matrix is already `[20.x, 22.x]`) and `package.json` has no `engines` field, so nothing in code contradicts it — that README line is the only statement of the floor anywhere in the repo.

That is a one-line README change and it belongs with #445 rather than in a competing lockfile PR. Happy to open it as a separate docs-only PR, or for it to just be folded into #445 — whichever is less work for you.

## Note on #454

#454 (blocking audit gate) needs a lockfile refresh to land first, and its body already names #445 as the prerequisite. Closing this changes nothing there — the ordering is still **#445, then #454**.

Issue #186, opened 2026-05-26
launch_tv_debug_mac.sh: pkill -f sends SIGTERM (trapped); single-instance lock kills the new --remote-debugging-port instance


hangocduong:
## Summary

On macOS, `scripts/launch_tv_debug_mac.sh` can leave you with TradingView running **without** the debug port, even after the script prints `DevTools listening on ws://127.0.0.1:9222/...`. The result is a confusing "Warning: CDP not responding after 15s" followed by a TradingView instance that quietly came back as the original (no-flag) parent.

### Root cause

`line 47` uses default `pkill -f "TradingView"` (SIGTERM). TradingView intercepts SIGTERM and the **single-instance lock survives**. The script then launches a fresh TV with `--remote-debugging-port=9222`; that new process briefly opens CDP (you see the "DevTools listening" log), but the still-alive original instance reaps it via the single-instance handoff and the port closes. Net effect: the new debug PID is gone, the original PID is still up, port 9222 is dead.

### Secondary issue

`"$APP" --remote-debugging-port=$PORT &` (line 52) has no `nohup`/`disown`, so the backgrounded child can receive SIGHUP when the calling shell exits in some launch contexts.

## Environment

- TradingView Desktop **3.1.0** (Electron 38.2.2)
- macOS **25.3.0** (Darwin), arm64
- Node v25.2.1
- Upstream `tradesdontlie/tradingview-mcp@main`

Distinct from #130: in my repro the wrapper *does* accept the `--remote-debugging-port` flag (CDP briefly binds). The breakage is purely the kill/relaunch race above.

## Reproduction

```bash
# TV Desktop already running normally (no debug flag)
./scripts/launch_tv_debug_mac.sh
# → "DevTools listening on ws://127.0.0.1:9222/..."
# → "Warning: CDP not responding after 15s"

pgrep -lf TradingView      # original parent PID still alive
lsof -iTCP:9222 -sTCP:LISTEN -P -n   # empty
```

Workaround that always works:

```bash
pkill -9 -f TradingView
sleep 2
nohup /Applications/TradingView.app/Contents/MacOS/TradingView \
  --remote-debugging-port=9222 >/tmp/tv-cdp.log 2>&1 &
disown
```

## Suggested fix

1. **SIGKILL + poll** instead of `pkill -f` + `sleep 1` — and fail loudly if processes are still alive after the timeout.
[...]

PR #463, opened 2026-08-17
fix(pine): stop pine_new/pine_smart_compile silently overwriting the open script

### LINKED ISSUE #395: pine_compile can silently overwrite the open saved script (clicks Save via textContent fallback)
## Summary

`pine_compile` / `pine_smart_compile` (and `tv pine compile`) can **silently overwrite the user's currently-open saved script** with whatever is in the editor buffer, creating a new cloud version of that script.

This happened to me on a real account: a 312-line saved strategy was replaced by a 13-line probe script as version 20. It was only recoverable because TradingView keeps version history.

## Root cause

`compile()` in `src/core/pine.js` selects the toolbar button by `textContent`:

```js
var text = btns[i].textContent.trim();
if (/save and add to chart/i.test(text)) { ... }
if (!fallback && /^(Add to chart|Update on chart)/i.test(text)) { fallback = btns[i]; }
if (!saveBtn && btns[i].className.indexOf('saveButton') !== -1 && btns[i].offsetParent !== null) { saveBtn = btns[i]; }
...
if (fallback) { fallback.click(); return fallback.textContent.trim(); }
if (saveBtn) { saveBtn.click(); return 'Pine Save'; }   // <-- destructive
```

In current TradingView Desktop the Pine Editor toolbar buttons are **icon-only** — `textContent` is `""`. So every regex above fails, `fallback` stays `null`, and control reaches the `saveBtn` branch, which **clicks Save**.

Observed result: `{"success": true, "button_clicked": "Pine Save", "has_errors": false, "study_added": false}` — reported as success, while the open saved script was overwritten and nothing was added to the chart.

Note the `Cmd+Enter` keyboard fallback below it is the *correct* action, but it is unreachable: it only runs `if (!clicked)`, and `saveBtn` matching makes `clicked` truthy.

## Reproduction

1. Open a **saved** script in the Pine Editor.
2. `pine_set_source` with different content.
3. `pine_smart_compile`.
4. The saved script now has a new cloud version containing the injected content.

## Why it is easy to miss

- The call is named "compile" and reads as non-destructive.
- It returns `success: true`.
- The damage is invisible from the editor — you have to re-fetch the saved version server-side to see it.

## Suggested fix

Target the stable QA hooks instead of `textContent`, and **delete the `saveButton` fallback entirely** — a compile path should never save:

```js
var add = document.querySelector('[data-qa-id="add-script-to-chart"]');
if (add && !add.disabled) { add.click(); return add.getAttribute('title'); }  // "Add to chart" / "Update on chart"
return null;   // let the Cmd+Enter fallback handle it
```

Relevant `data-qa-id` values in the current UI:

| Element | Selector |
| --- | --- |
| Add / Update on chart | `button[data-qa-id="add-script-to-chart"]` (hotkey ⌘+↵) |
| Save (never click implicitly) | `button[data-qa-id="pine-script-save-button"]` |
| Script title + menu | `div[role=button][data-qa-id="pine-script-title-button"]` |
| Publish | `button[data-qa-id="publish-script"]` |
| More options | `button[data-qa-id="script-more-options"]` |

Two suggestions beyond the direct fix:

1. **Add a pre-write guard.** Before injecting source, read the open script's name via `[data-qa-id="pine-script-title-button"]`, and refuse (or require an explicit `force`) when it is a saved script rather than `"Untitled script"`.
2. **Note in the README** that a brand-new untitled script can be added to the chart *without* saving, but a modified saved script cannot — TradingView blocks it with a "Save this script before adding?" dialog. That dialog is the reason a save is being attempted at all, so callers need a documented safe path.

## Related (separate bug, same area)

`pine_get_source` / `pine_set_source` frequently act on the **wrong Monaco instance**. There are two `.monaco-editor.pine-editor-monaco` elements; `FIND_MONACO` uses `querySelector` (first match), which is a hidden editor with `getBoundingClientRect().width === 0`. Symptoms: `pine_get_source` returns 0 chars while the editor visibly has content, and `pine_set_source` reports success while the visible editor is unchanged. Selecting by visible bounding box fixes it:

```js
var eds = env.editor.getEditors();
for (var i = 0; i < eds.length; i++) {
  var dn = eds[i].getDomNode();
  if (dn && dn.getBoundingClientRect().width > 0) return eds[i];
}
```

The fiber walk also gives up at 15 `.return` levels and only tries the first container; widening both made it reliable for me.

## Environment

- tradingview-mcp @ `0ac960a`
- TradingView Desktop 3.3.0 (Electron 38.2.2, Chrome 140.0.7339.133)
- macOS (Darwin 25.4.0), Node v23.9.0

EntryCall:
Fixes #395. Related: #352, #158.

## Problem

`pine_new` → `pine_set_source` → `pine_smart_compile` overwrites whichever saved script happens to be open in the Pine Editor. I lost a script to this today; #395 reports the same thing ("a 312-line saved strategy replaced by a 13-line probe script"). Two separate defects combine:

### 1. `smartCompile()` falls through to Save

```js
if (!addBtn && /^add to chart$/i.test(text)) addBtn = btns[i];
...
if (!saveBtn && btns[i].className.indexOf('saveButton') !== -1 ...) saveBtn = btns[i];
```

`addBtn`/`updateBtn` are matched on `textContent`, but TradingView Desktop 3.x renders them **icon-only** — `textContent` is empty, so neither matches. `saveBtn` is matched on **className**, which always matches. So every compile clicked Save, persisting the buffer into the bound script slot.

Verified live against TradingView Desktop 3.3.0 by running both matchers against the real DOM:

```
OLD_would_click: "Pine Save (DESTRUCTIVE)"
NEW_would_click: "Add to chart"
addBtn_label:    "Add to chart"   <- via aria-label; textContent is ""
```

The accessible name is on `aria-label`, not `textContent`. That is the whole bug.

### 2. `newScript()` never created a script

```js
// Simply set the source to a new template — this is the most reliable approach
m.editor.setValue(template);
return { success: true, type, action: 'new_script_created', ... };
```

It only replaced the buffer text, leaving it bound to the previously open script, then reported `new_script_created`. Combined with (1), the next compile wrote the new source over the user's existing script.

## Fix

**`smartCompile({ allowSave = false })`**
- Fold `aria-label` / `title` / `data-name` / `class` into the button match so the non-destructive buttons are found again on icon-only builds.
- Gate every Save path behind explicit opt-in (`allow_save`, default `false`). With no non-destructive button available it now falls through to Ctrl+Enter rather than Save.
- Scope the Save lookup to the Pine Editor panel — the **chart layout's** save button also carries the `saveButton` class prefix and was reachable from the old selector (confirmed: it resolved to `"UnnamedSave All changes saved"`).

**`newScript({ type })`**
- Drives the script-name dropdown's documented **Create new → indicator/strategy/library** ([TradingView docs](https://www.tradingview.com/support/solutions/43000711497-how-to-create-new-script/)), falling back to the "More" menu.
- Re-reads the editor's binding afterwards; if it still shows a saved version stamp, it **throws** instead of returning success.
- Only ever clicks a menu item matching the `new blank <type>` pattern — that menu also holds destructive items.
- There is no longer any code path where `newScript` mutates the open script's contents. It fails loudly instead.

DOM lookups are anchored to `.tv-script-widget`; sibling classes in that subtree are build-hashed (`editorWrapper-mImut1T6`, `container-FgScb_09`) and unsafe to match on.

## Compatibility

`allow_save` is optional and defaults to false, so this is a behaviour change for anyone who relied on `pine_smart_compile` implicitly saving. That seems clearly right given the failure mode, but happy to add a `pine_compile_and_save` alias instead if you'd prefer to keep the old default reachable.

## Testing

- `node --check` on both files; module imports clean.
- Both matchers exercised against a live TradingView Desktop 3.3.0 session via CDP (results above).
- The `newScript` menu selectors were validated read-only against the live DOM (`button[class*="nameButton"]` resolves to the script-name dropdown; `button[class*="saveButton"]` scoped to `.tv-script-widget` correctly excludes the chart-layout button).

I could not exercise the full `newScript` click-through end-to-end without risking another live script, so the menu-item click path is verified by selector resolution rather than a completed run. Worth a second pair of eyes there.

### DIFF
--- a/src/core/pine.js
+++ b/src/core/pine.js
@@ -565,10 +565,12 @@ export async function newScript({ type }) {
   if (!editorReady) throw new Error('Could not open Pine Editor.');
 
   const typeMap = { indicator: 'indicator', strategy: 'strategy', library: 'library' };
+  // Submenu labels carry their shortcut inline ("Indicator⌘ K, ⌘ I"), so anchor
+  // at the start only. "Built-in…" sits in the same submenu and must not match.
   const patterns = {
-    indicator: 'new blank indicator|new indicator',
-    strategy: 'new blank strategy|new strategy',
-    library: 'new blank library|new library',
+    indicator: '^indicator',
+    strategy: '^strategy',
+    library: '^library',
   };
   const wanted = patterns[type] || patterns.indicator;
 
@@ -583,10 +585,13 @@ export async function newScript({ type }) {
     (function() {
       var root = document.querySelector('${PINE_ROOT}');
       if (!root) return false;
-      // The script-name button is the dropdown holding Create new / Make a copy /
-      // Version history. Class suffixes are build-hashed, so match on the prefix.
-      var title = root.querySelector('button[class*="nameButton"]');
-      if (title && title.offsetParent !== null) { title.click(); return 'title-dropdown'; }
+      // The script-name control is the dropdown holding Create new / Make a copy /
+      // Version history. It is a DIV[role=button], NOT a <button>, so do not
+      // restrict by tag. Class suffixes are build-hashed; match on the prefix.
+      var title = Array.prototype.slice.call(
+        root.querySelectorAll('[class*="nameButton"]')
+      ).filter(function(e) { return e.offsetParent !== null; })[0];
+      if (title) { title.click(); return 'title-dropdown'; }
       var more = Array.prototype.slice.call(
         root.querySelectorAll('button[aria-label="More"], button[data-name*="menu"], button[aria-label*="menu" i]')
       ).filter(function(b) { return b.offsetParent !== null; });
@@ -604,28 +609,39 @@ export async function newScript({ type }) {
 
   await new Promise(r => setTimeout(r, 400));
 
-  // Click the "New blank <type>" item. Only ever click an item matching this
-  // regex — never anything else in that menu (it also contains destructive items).
-  const itemClicked = await evaluate(`
+  // "Create new" opens a submenu; the type lives one level down. Only ever click
+  // items matching these exact patterns — the same menu holds destructive entries.
+  const MENU_SEL = '[role="menuitem"], [class*="item-"], [class*="label-"]';
+  const clickMenuItem = (pattern) => evaluate(`
     (function() {
-      var re = new RegExp(${JSON.stringify(wanted)}, 'i');
-      var nodes = document.querySelectorAll('[role="menuitem"], [class*="item" i], [class*="label" i]');
+      var re = new RegExp(${JSON.stringify('PLACEHOLDER')}, 'i');
+      var nodes = document.querySelectorAll('${MENU_SEL}');
       for (var i = 0; i < nodes.length; i++) {
         var n = nodes[i];
         if (n.offsetParent === null) continue;
         var t = (n.textContent || '').trim();
-        if (t.length > 60) continue;
+        if (t.length > 40) continue;
         if (re.test(t)) { n.click(); return t; }
       }
       return null;
     })()
-  `);
+  `.replace(JSON.stringify('PLACEHOLDER'), JSON.stringify(pattern)));
+
+  const submenuOpened = await clickMenuItem('^create new$');
+  if (!submenuOpened) {
+    await evaluate(`(function(){ document.body.click(); return true; })()`);
+    throw new Error('Could not find "Create new" in the Pine Editor script menu. Refusing to fall back to overwriting the open script.');
+  }
+
+  await new Promise(r => setTimeout(r, 400));
+
+  const itemClicked = await clickMenuItem(wanted);
 
   if (!itemClicked) {
     // Close the menu so we do not leave the UI in a half-open state.
     await evaluate(`(function(){ document.body.click(); return true; })()`);
     throw new Error(
-      'Could not find a "New blank ' + type + '" item in the Pine Editor menu. ' +
+      'Could not find a "' + type + '" item in the Pine Editor "Create new" submenu. ' +
       'Refusing to fall back to overwriting the open script.'
     );
   }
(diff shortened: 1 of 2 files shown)


EntryCall:
Pushed a correction (433abb8). I validated the selectors against a live TradingView Desktop 3.3.0 session and my first pass had guessed the `newScript` DOM wrong in three places:

1. **The script-name control is a `DIV[role="button"]`, not a `<button>.`** `root.querySelector('button[class*="nameButton"]')` matched nothing, so the menu never opened and `newScript` would have thrown on every call. Now matches on the class prefix only.

2. **There is no "New blank indicator" item.** The actual menu is:

   ```
   Save script ⌘S / Make a copy… / Rename… / Version history… /
   Move script to bottom / Create new / Open script… ⌘O
   ```

   and `Create new` opens a **submenu**: `Indicator⌘ K, ⌘ I` / `Strategy⌘ K, ⌘ S` / `Library` / `Built-in…`. `newScript` now walks both levels.

3. **Submenu labels carry the shortcut inline**, so the type patterns anchor at the start only. Verified `^indicator` matches `Indicator⌘ K, ⌘ I` and does *not* match the adjacent `Built-in…`.

Also switched the menu-item lookup to `[class*="item-"]` / `[class*="label-"]` (hyphen included) so it binds to the hashed CSS-module classes rather than any element whose class merely contains "item".

The `smartCompile` half is unchanged and was already verified live — old matcher selects `Pine Save`, new one selects `Add to chart`.

Separately, on #186: `pkill -f TradingView` did not kill the app for me either. The main process survived and the relaunch got absorbed by the single-instance lock, coming back up with **no** `--remote-debugging-port` (silent — `tv_launch` still returned a pid and `success: true`). A hard `kill -9` on every `TradingView.app` pid, confirming 0 remaining, then relaunching was the only thing that worked. Happy to send that as a separate PR if useful.

PR #493, opened 2026-08-28
Add local historical data tools and Trend Join Long day-trading scanner

afberrezueta:
## Summary

- **`historical_list_symbols` / `historical_get_ohlcv`** — serve local 10-year daily OHLCV CSVs (Financial Modeling Prep) independent of the live CDP connection. Useful for longer-range stats/backtesting beyond the live chart's 500-bar cap, or when TradingView isn't running. Symbol resolution goes through an on-disk allowlist rather than joining raw input into a path, to avoid path traversal via a crafted `symbol` argument. Underlying CSVs live in `data/` (gitignored — user-specific, not shipped in the repo).

- **`scanner_run_tjl`** — sequential per-ticker Trend Join Long breakout scanner over a user-editable watchlist (`data/tjl_watchlist.json`, gitignored — flip `enabled` or add tickers, no restart needed). Gates on a live CDP connection and on 10:00am-3:30pm New York market hours, saving a graceful error JSON when the time gate is closed instead of scanning outside the intended window. The 1-min OHLCV pull is capped at 500 bars (the live-chart hard cap) with a self-reported `data_gap` warning when that window can't reach back to the 4:00am premarket start, rather than silently returning a wrong/null premarket high.

- Updates the tool count (84 → 87) across `CLAUDE.md`, `README.md`, and the server's MCP instructions string, plus decision-tree entries for both new capabilities.

## Test plan

- [x] `npm run lint` — clean (only 3 pre-existing warnings, none in new files)
- [x] `npm run test:unit` — 154/154 passing, including the source-audit test that covers the new `historical.js` for unsafe `evaluate()` interpolation
- [x] Server boots cleanly with both new tool groups registered
- [x] `historical_get_ohlcv` / `historical_list_symbols` exercised directly (symbol lookup, count/date-range filtering, unknown-symbol error path, path-traversal attempt correctly rejected)
- [x] `scanner_run_tjl` exercised live against the real TradingView CDP connection: sequential multi-ticker run, time-gate closed → error JSON saved correctly, `data_gap` warning fires correctly late in the day

## Notes for reviewers

- Both new tools follow the existing `core/` + `tools/` + `server.js` registration pattern used throughout the codebase.
- No new dependencies added.


afberrezueta:
Closing for now — decided to keep this feature set (historical data tools + TJL scanner) proprietary rather than contribute it upstream. Thanks for maintaining this project, and apologies for the noise.

Issue #470, opened 2026-08-20
MSIX copy-fallback crashes immediately (exit code 9) on TradingView 3.3.0.7992 / Win11 — CDP never binds


ArgoKingo:
Body:

## Environment
- TradingView Desktop: `3.3.0.7992` (MSIX)
  - PackageFullName: `TradingView.Desktop_3.3.0.7992_x64__n534cwy3pjxzj`
  - PackageFamilyName: `TradingView.Desktop_n534cwy3pjxzj` (no publisher prefix, so this isn't the #366 detection bug)
- OS: Windows 11 Home, 10.0.26200
- tradingview-mcp: latest `main` (installed 2026-08-20)

## What I did
1. `scripts\launch_tv_debug.bat` (and separately, the equivalent of `src/core/health.js`'s `launch()`): direct spawn of
   `C:\Program Files\WindowsApps\TradingView.Desktop_3.3.0.7992_x64__n534cwy3pjxzj\TradingView.exe --remote-debugging-port=9222`
   → **No process appears at all** (`Get-Process` shows nothing), no CDP, batch script just loops on "Still waiting..." forever.
2. Manually reproduced the local-copy fallback from `_copyMsixPackageLocal`:
   ```powershell
   $pkg = (Get-AppxPackage TradingView.Desktop).InstallLocation
   Copy-Item "$pkg\*" "$env:LOCALAPPDATA\tradingview-mcp\TradingView" -Recurse -Force
   & "$env:LOCALAPPDATA\tradingview-mcp\TradingView\TradingView.exe" --remote-debugging-port=9222
Copy is byte-identical (211,185,664 bytes, same 26 top-level entries, no reparse points lost). The process does start this time (confirmed via Start-Process -PassThru), but exits within ~3 seconds with ExitCode 9, before CDP ever has a chance to bind. No crash dialog, nothing in Application event log, no Defender detection.

Why this seems new
[...]

Issue #393, opened 2026-07-26
Windows MSIX: tv_launch fails with spawn EPERM; copy-fallback crashes and re-register can break existing install


JAKFAR999:
**OS**: Windows 10 Pro (10.0.19045), TradingView Desktop 3.3.0.7992 (x64), installed via Microsoft Store (MSIX, under WindowsApps)
**What I tried**
1. `tv_launch` (MCP tool):
```json
{"success": false, "error": "spawn EPERM"}
2. Manual scripts\launch_tv_debug.bat — found the exe correctly, but launching it directly failed:
Found TradingView at: C:\Program Files\WindowsApps\TradingView.Desktop_3.3.0.7992_x64__n534cwy3pjxzj\TradingView.exe
Starting with --remote-debugging-port=9222...
Access is denied.
...
Error: TradingView is running but CDP never became available on port 9222.
3. Manual copy-fallback from SETUP_GUIDE.md (Copy-Item "$pkg\*" "$env:LOCALAPPDATA\tradingview-mcp\TradingView" -Recurse -Force, then launch from the copy):
The copied exe launched and CDP responded briefly on /json/version, but the app then crashed with an in-app error dialog (Restart/Exit options) before any chart page loaded. /json (target list) timed out during this window.
Tried re-registering the copy as a dev-sideloaded package (enabled Windows Developer Mode, granted ALL APPLICATION PACKAGES / ALL RESTRICTED APPLICATION PACKAGES read+execute ACLs on the copy folder, then Add-AppxPackage -Register) to preserve package identity. Registration still failed:
Deployment failed with HRESULT: 0x80073CF0, Package could not be opened.
error 0x80070003: Opening file from location: AppxManifest.xml failed with error: The system cannot find the path specified.
The original install folder contains a microsoft.system.package.metadata directory, suggesting this is a Store-optimized/hardlinked package that can't be relocated or re-registered from a plain file copy.
Heads up for other Windows Store users: the failed Add-AppxPackage -Register attempt above actually unregistered the original package for the current user as a side effect (files stayed on disk under WindowsApps, but Get-AppxPackage no longer listed it, and TradingView could no longer be launched from the Start Menu at all). Recovered by re-registering the original manifest in place:

Add-AppxPackage -Register "C:\Program Files\WindowsApps\TradingView.Desktop_<version>\AppxManifest.xml" -DisableDevelopmentMode
[...]

Issue #348, opened 2026-07-12
MSIX/Store install: CDP launch fails with Access Denied (direct) and STATUS_BREAKPOINT self-terminate (local copy fallback)


mdaleem01-lang:
**Environment:** Windows 10, TradingView.Desktop_3.3.0.7992_x64 (Microsoft Store / MSIX)

**What happened:**
1. Direct launch of TradingView.exe from the WindowsApps install location with
   `--remote-debugging-port=9222` fails immediately with "Access is denied."
2. Tried the documented local-copy fallback (copy the package folder to
   %LOCALAPPDATA% and launch from there). The copy succeeds, but the process
   exits immediately with exit code -2147483645 (0x80000003, STATUS_BREAKPOINT)
   -- looks like an identity/license check killing the process when it's not
   running inside its real package context.
3. Re-downloaded TradingView.msix directly from
   https://tvd-packages.tradingview.com/stable/latest/win32/TradingView.msix
   and sideloaded it -- it installs as the exact same PackageFullName
   (TradingView.Desktop_3.3.0.7992_x64__n534cwy3pjxzj) and hits the identical
   Access Denied error, confirming there's no separate non-Store distribution
   to fall back to on Windows.

**Expected:** tv_launch (or the manual launch scripts) successfully opens
CDP on port 9222.

**Actual:** Both the direct launch and the local-copy fallback fail as above.

Possibly related to #42, #75, #128.

saagarverma1234:
Got CDP working on this exact setup. Posting the approach in case it helps others in this thread — with the caveat up front that this is a single machine and one build, so I can't claim it generalizes.

**TL;DR:** `Invoke-CommandInDesktopPackage` **without** `-PreventBreakaway` launches the app inside its MSIX container *and* passes `--remote-debugging-port` through. No admin rights, no Developer Mode, no local copy.

```powershell
$pkg = Get-AppxPackage | Where-Object { $_.Name -like '*TradingView*' } | Select-Object -First 1
$exe = Join-Path $pkg.InstallLocation 'TradingView.exe'
$appId = @((Get-AppxPackageManifest $pkg.PackageFullName).Package.Applications.Application)[0].Id

Invoke-CommandInDesktopPackage `
  -PackageFamilyName $pkg.PackageFamilyName `
  -AppId $appId `
  -Command $exe `
  -Args "--remote-debugging-port=9222"
```

Result:

```json
{
  "Browser": "Chrome/140.0.7339.133",
[...]

PR #503, opened 2026-09-04
fix(launch): route synchronous spawn EPERM into MSIX local-copy fallback

kubilaykoray-collab:
On some Windows builds, spawn() on a WindowsApps-packaged TradingView.exe throws EPERM synchronously instead of emitting an 'error' event. The call to _spawnDetached sat outside any try/catch, so the exception escaped launch() before the local-copy fallback could run — tv_launch failed with a bare "spawn EPERM" and the fallback was effectively dead code for that failure mode.

Wrap the initial spawn and treat a synchronous throw as an early failure, so both the sync-throw and async-'error' paths reach the same fallback. Non-MSIX installs rethrow unchanged.

The existing EACCES test only mocked the async 'error' event, which is why this went unnoticed. Adds a regression test for the synchronous throw (verified failing without the fix) plus a guard test that a classic install still propagates a spawn error rather than copying.

Claude-Session: https://claude.ai/code/session_01Pzp65QL62bkuw6xaUm7yeg

### DIFF
--- a/src/core/health.js
+++ b/src/core/health.js
@@ -360,12 +360,25 @@ export async function launch({ port, kill_existing, _deps } = {}) {
   if (killFirst) await killExisting();
 
   const cdpArgs = [`--remote-debugging-port=${cdpPort}`];
-  let child = _spawnDetached(deps.spawn, tvPath, cdpArgs);
+  const isWindowsApps = platform === 'win32' && WINDOWS_APPS_RE.test(tvPath);
+
+  // Blocked MSIX launches can surface either way: spawn throws EPERM synchronously
+  // on some builds, or emits 'error'/exits on others. Capture the sync throw so both
+  // paths reach the local-copy fallback below.
+  let child = null;
+  let spawnThrew = null;
+  try {
+    child = _spawnDetached(deps.spawn, tvPath, cdpArgs);
+  } catch (e) {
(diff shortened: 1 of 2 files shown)


d4rkwinz:
Verified on a live MSIX install — works.

**Env:** Win10 Pro 19045 · TradingView Desktop 3.4.1.8194 (MSIX, `TradingView.Desktop_3.4.1.8194_x64__n534cwy3pjxzj`) · Node 24.19.0 · repo at c05b8f5

**Before:** `tv launch` → `{"success": false, "error": "spawn EPERM"}`. `launch_tv_debug.bat` doesn't reach the fallback either — `start` gives `Access is denied.`, then 30 polls and exit 1.

**After:** `{"success": true, "msix_local_copy": true, "browser": "Chrome/146.0.7680.216"}` in 3.4s. `tv status` → `cdp_connected: true`, `api_available: true`, session and layout intact.

**Re `--no-sandbox` (#529 bug 2, #530):** #529 is the same build as mine and says the local copy needs `--no-sandbox --disable-gpu` or it dies with "GPU process isn't usable". Didn't reproduce here — copy ran with only `--remote-debugging-port=9222`: 10 processes, same `target_id`, CDP still up ~6 min later. The crash reports I found (#470, #348, #393) are all Win11, so the sandbox flag may be separable from this fix.

PR #505, opened 2026-09-04
fix(pine): interpolate pine_check error messages and annotate the source line

Bern-code:
## The bug

`pine_check` returns TradingView's message templates verbatim, so a real compile error reads:

```
Undeclared identifier "{identifier}"
```

pine-facade doesn't interpolate its own messages. It sends the values separately in a `ctx` map next to the template, and `check()` in `src/core/pine.js` was reading `e.message` while dropping `e.ctx`:

```json
{"code": "CE10272",
 "ctx": {"identifier": "risk"},
 "message": "Undeclared identifier \"{identifier}\"",
 "start": {"line": 139, "column": 5}}
```

So the one field the caller actually needs — which identifier — was being thrown away on every error.

## The fix

Fill the placeholders from `ctx`, pass the error `code` through, and add an `annotated` view showing the offending line with a caret under the reported column:

```
 139 |     risk := atrVal * slMult
     |     ^-- Undeclared identifier "risk"
```

A placeholder with no matching `ctx` key is left as-is rather than blanked, so an unrecognised template degrades to today's behaviour instead of losing text. Warnings get the same treatment. `annotated` is omitted entirely when the script is clean, so nothing changes for a passing check.

This matters most on builds where the Pine editor has no React fiber — `pine_set_source` / `pine_smart_compile` / `pine_get_errors` all fail there (several open PRs are chasing variants of this), and `pine_check` becomes the only compile path left. Its output is then the entire error report.

## What I ran

```
npm run lint       # 0 errors (4 pre-existing warnings, untouched)
npm run test:unit  # 160/160 pass
```

Against the live pine-facade endpoint, 15 real third-party Pine scripts (v5 and v6, 2.5KB–20KB) were checked end to end: 13 compiled clean, 2 had genuine errors. Both error cases printed raw `{identifier}` / `{funId}` templates before this change and read correctly after:

```
 139 |     risk := atrVal * slMult
     |     ^-- Undeclared identifier "risk"

  87 |     plot(smaFast * (1 - t) + smaSlow * t, title="Band " + str.tostring(i),
     |                                                 ^-- Cannot call "plot" with argument "title"="call "operator +" (series string)". An argument of "series string" type was used but a "const string"  is expected.
```

The second one is the case that convinced me this was worth filing — untouched, it names neither the function nor the argument.

`npm run test:e2e` was **not** run: no TradingView Desktop on this machine, and this path never opens a CDP connection. `check()` posts to pine-facade from the server process, so there is nothing browser-side for the change to affect.

## Tests

`tests/pine_annotate.test.js`, 8 tests, fetch stubbed — no network, no TradingView. Covers caret column alignment, out-of-range and missing-column input, multiple issues, `ctx` interpolation for errors and warnings, an unknown placeholder, and `annotated` being absent on a clean compile.

https://claude.ai/code/session_01DsdBoQLyPJihjuyYbPkda6

### DIFF
--- a/CLAUDE.md
+++ b/CLAUDE.md
@@ -42,6 +42,17 @@ Use `study_filter` parameter to target a specific indicator by name substring (e
 - `chart_set_visible_range` → zoom to exact date range (unix timestamps)
 
 ### "Work on Pine Script"
+0. `pine_check` → **always start here.** Compiles against TradingView's own
+   server-side compiler without opening the editor or needing a chart. Returns
+   `errors` / `warnings` with line+column+code, plus an `annotated` view:
+
+   ```
+    139 |     risk := atrVal * slMult
+        |     ^-- Undeclared identifier "risk"
+   ```
+
+   Fix everything it reports before touching the editor. The chart is where you
+   look at an idea, not where you debug it.
 1. `pine_set_source` → inject code into editor
 2. `pine_smart_compile` → compile with auto-detection + error check
 3. `pine_get_errors` → read compilation errors
@@ -50,6 +61,10 @@ Use `study_filter` parameter to target a specific indicator by name substring (e
 6. `pine_save` → save to TradingView cloud
 7. `pine_new` → create blank indicator/strategy/library
 8. `pine_open` → load a saved script by name
+
+Steps 1-8 drive the Monaco editor through the React fiber tree, which does not
+exist on every TradingView build. `pine_check` and `pine_analyze` never touch
+the browser, so they work regardless.
 
 ### "Practice trading with replay"
 1. `replay_start` with `date: "2025-03-01"` → enter replay mode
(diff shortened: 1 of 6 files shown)


Bern-code:
Both Copilot points addressed in `8e17be2`:

- **CRLF** — `annotate()` now splits on `/\r?\n/`. A CRLF script was leaving a stray `\r` in the rendered line, which pushed the caret out of alignment with the text above it.
- **Column clamp** — the column is now required to be an integer and clamped before it reaches `repeat()`.

One nuance on the clamp: the upper bound is `line.length + 1`, not `line.length`. `Syntax error at input "end of line without line continuation"` legitimately points one character past the end, and clamping to the text would have moved that caret onto the last character and made it read as if the problem were there. So the clamp kills the huge-allocation case without changing any real compiler output.

Three tests added for it — CRLF alignment, a `column: 9_999_999` that stays under 200 chars, and the end-of-line caret still landing at `src.length`. 163/163 unit tests pass, lint clean.

PR #506, opened 2026-09-04
feat: optional JSONL log of every tool call

Bern-code:
## What

Set `TV_MCP_LOG_FILE` and every tool call is appended as one JSON line — name, arguments, truncated result, duration:

```json
{"ts":"2026-09-04T12:30:08.145Z","tool":"pine_check","args":{"source":"plot(close)"},"duration_ms":312.4,"result":{"is_error":false,"chars":35,"text":"..."}}
```

## Why

Debugging this bridge means reconstructing what an agent did to a live chart, and right now there's no record. A stray `ui_click` that closes the Pine panel, a `chart_set_symbol` nobody meant to issue, an indicator whose settings dialog got opened — all of it is invisible after the fact. With a log it's one `grep`.

It also answers "which tool is slow on this build", which varies a lot between Desktop versions:

```bash
jq -r 'select(.result.is_error) | "\(.ts) \(.tool)"' calls.jsonl
jq -s 'group_by(.tool) | map({tool: .[0].tool, n: length,
       p95: (sort_by(.duration_ms) | .[(length * 0.95 | floor)].duration_ms)})' calls.jsonl
```

And it makes bug reports concrete — `CONTRIBUTING.md` asks for "the exact tool call with its full JSON result", which is exactly a line of this file.

## How

One wrapper around `server.tool()` in `server.js`. All 84 tools are covered without touching any `register*()` module, and new tools are logged for free.

Deliberately cheap and opt-in:

- **Unset (the default), `instrument()` returns the server untouched** — no wrapper on the hot path, not even a branch per call
- Results cap at 2000 chars so screenshot payloads don't bloat the file
- A failing write is swallowed — logging must never break a tool call
- A thrown handler is logged with `threw` and rethrown unchanged

Nothing is written unless the user opts in, and the file stays on their machine. No telemetry, no network, nothing added to the default path.

## Scope check

Per `CONTRIBUTING.md`: no server access, no market data stored (results are truncated tool output, not price series), no auth bypass, no execution. It's local debug tooling for the bridge itself.

## What I ran

```
npm run lint       # 0 errors (4 pre-existing warnings, untouched)
npm run test:unit  # 160/160 pass
```

Server boot with the env var set and unset, confirming `tools/list` still returns all 84 tools either way, and that `tools/list` alone writes no lines (it isn't a tool call):

```
$ echo '{"jsonrpc":"2.0","id":1,"method":"tools/list","params":{}}' | TV_MCP_LOG_FILE=/tmp/calls.jsonl node src/server.js
⚠  tradingview-mcp  |  Unofficial tool. Not affiliated with TradingView Inc. or Anthropic.
   Ensure your usage complies with TradingView's Terms of Use.
   Tool-call log: /tmp/calls.jsonl
```

`npm run test:e2e` was **not** run: no TradingView Desktop on this machine. The wrapper is transport-level and never touches CDP — it wraps the handler the SDK already calls, and passes its arguments and return value through untouched.

## Tests

`tests/tool_log.test.js`, 6 tests, offline. They spawn child processes because `TV_MCP_LOG_FILE` is read at module load. Covers the written line's shape, no file when unset, a thrown handler being logged and rethrown, truncation of a 50KB result, directory creation, and tools registering normally with logging off.

https://claude.ai/code/session_01DsdBoQLyPJihjuyYbPkda6

### DIFF
--- a/CLAUDE.md
+++ b/CLAUDE.md
@@ -120,6 +120,20 @@ These tools can return large payloads. Follow these rules to avoid context bloat
 - OHLCV capped at 500 bars, trades at 20 per request
 - Pine labels capped at 50 per study by default (pass `max_labels` to override)
 
+## Tool-Call Logging
+
+Set `TV_MCP_LOG_FILE` in the server's `env` to append every tool call to a JSONL
+file — name, args, truncated result, duration. Unset it is a no-op. Use it when
+a chart ends up in a state nobody meant to create, or to time a slow tool.
+
+```bash
+jq -r 'select(.result.is_error) | "\(.ts) \(.tool)"' ~/.tradingview-mcp/calls.jsonl
+jq -s 'group_by(.tool) | map({tool: .[0].tool, n: length,
+       p95: (sort_by(.duration_ms) | .[(length * 0.95 | floor)].duration_ms)})' calls.jsonl
+```
+
+Results are truncated at 2000 chars so screenshots do not bloat the file.
+
 ## Architecture
 
 ```
--- a/CLAUDE.md
+++ b/CLAUDE.md
@@ -132,7 +132,8 @@ jq -s 'group_by(.tool) | map({tool: .[0].tool, n: length,
        p95: (sort_by(.duration_ms) | .[(length * 0.95 | floor)].duration_ms)})' calls.jsonl
 ```
 
-Results are truncated at 2000 chars so screenshots do not bloat the file.
+Arguments and results are truncated at 2000 chars per string, so a pasted Pine
+script or a screenshot payload does not bloat the file.
 
 ## Architecture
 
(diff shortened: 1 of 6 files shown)


Bern-code:
Both Copilot points addressed:

- **Argument size** — this was the real one, thanks. Results were capped but args went in verbatim, so a single `pine_set_source` call could drop a 200KB script into the log. Long strings are now truncated the same way as results, at any depth in the argument object, with the original length recorded (`…[truncated 200000 chars]`). The object's shape is kept, so the log still shows *which* arguments were passed — just not all their bytes. Non-string values pass through untouched.
- **`fileURLToPath`** — fixed, now matching `tests/cli.test.js` and `src/core/update.js`.

Added a test covering a 200KB `source`, a nested long string, and a boolean passing through unchanged. 161/161 unit tests pass, lint clean.

Docs updated in both places to say arguments and results are capped, not just results.

PR #522, opened 2026-09-14
fix(pine): detect icon-only add/update-to-chart button

alexisiandriotis-cmd:
## Problem

`pine_compile` and `pine_smart_compile` report success but never add the script to the chart — `study_added` stays `false`, with no error surfaced.

Both matched the add-to-chart control by button **text**:

```js
if (!addBtn && /^add to chart$/i.test(text)) addBtn = btns[i];
```

On TradingView Desktop 3.4.x that control is an **icon-only button**: no text content, and no `title` attribute until a script has been added once (after which it gains `title="Update on chart"`). The text match never hits, so both functions fall through to the `saveButton` branch and click **Save** instead.

The failure is silent — the tool returns `{ success: true, button_clicked: "Pine Save" }` while the study never reaches the chart.

## Fix

Extract the shared lookup into a `FIND_RUN_BUTTON` constant, alongside the existing `FIND_MONACO`, and resolve the button in order:

1. `save and add to chart` text (unchanged)
2. `title` / `aria-label` — catches the post-add `Update on chart` state
3. button text (previous behaviour, retained)
4. the button's **SVG path prefix** — catches a fresh script that has no title yet
5. Save button, last resort only

This also de-duplicates two near-identical blocks that had already drifted apart (one used `/^(Add to chart|Update on chart)/i`, the other `/^add to chart$/i` plus a separate update branch).

Additionally: replaced the `offsetParent !== null` visibility test with `getBoundingClientRect()`, since `offsetParent` is `null` for `position: fixed` elements and the Pine Editor renders as a floating dialog.

## Verification

Against TradingView Desktop **3.4.1** (Electron 41.7.1, Chrome 146), with the editor open and a script already added:

| Branch | Result |
|---|---|
| `byTitle` (new) | ✅ `Update on chart` |
| `byText` (previous logic) | ❌ `null` — the bug |
| `saveBtn` (previous fallback) | ⚠️ `Save script` — what got clicked |

The `byPath` branch was confirmed separately against a freshly created script, where the button carries no title and the icon is the ▷ play glyph; clicking the element matched that way added the study successfully. The icon swaps to a refresh glyph after the first add, which is why both branches are needed.

`npm run test:unit` — 152/152 pass. ESLint clean.


alexisiandriotis-cmd:
Follow-up: confirmed end-to-end through `pine.compile()` itself, not just the selector logic.

Imported the patched module directly against a live TradingView Desktop **3.4.1** (Electron 41.7.1) session on CDP port 9222, created a fresh script, injected source, and called `compile()`:

```
1. pine.newScript()  -> {"success":true,"action":"new_script_created"}
2. pine.setSource()  -> {"success":true,"lines_set":3}
3. pine.compile()    -> {"success":true,"button_clicked":"Update on chart"}

studies before : 6
newly added    : ["ZZ Patch Probe"]
RESULT: PASS — compile() put the study on the chart
```

Pre-patch, the same call returned `button_clicked: "Pine Save"` and the study never reached the chart.

Scope of what this run covers: the button carried `title="Update on chart"`, so the title branch resolved first and the SVG-path branch was not reached. The path branch is what handles a fresh script that has no title yet — verified separately by clicking the element matched that way, which added the study successfully, but not through `compile()` in that state.

One incidental note for anyone testing similarly: when the editor is still bound to an on-chart study, `compile()` clicks **Update on chart** and replaces that study in place rather than adding a sibling. Expected behaviour, but easy to lose an indicator to if you are probing against a populated chart.

PR #536, opened 2026-09-20
fix(drawing): accept color/linewidth/linestyle on draw_shape, report what landed

joebowen1208:
### Problem

`draw_shape` accepted style only via `overrides` (a JSON string). Zod object schemas strip unknown keys by default, so a caller passing `color`, `linecolor`, `linewidth`, `linestyle`, or a `properties` object had every one silently discarded — and still got `success: true` back.

The shape drew with TradingView's defaults (yellow, width 1, solid) and nothing in the response indicated anything had been dropped. I hit this plotting 31 price levels with a per-label colour palette; all 31 came back `success: true` and all 31 were default yellow. It was only caught by eyeballing a screenshot, then confirming via `draw_get_properties` that `linecolor` was `rgba(255, 235, 59, 0.9809)`.

The `overrides` path itself works correctly — I verified that separately. The defect is entirely in **what the schema accepts** and in **the absence of any feedback when a style doesn't land**.

### Changes

- Accept `color` / `linecolor` / `linewidth` / `linestyle` / `textcolor` / `fontsize` as first-class params, mapped into overrides. Explicit `overrides` still wins on conflict, so it remains the lower-level escape hatch.
- Accept `overrides` as an object as well as a JSON string.
- Read the style back off the chart after creation and return `style_requested` / `style_applied`, plus `style_not_applied` listing any key that didn't take. Colours compare through a normalizer, so `#FF80AB` and `rgba(255,128,171,0.98)` are equal but TradingView's default yellow is not.
- The read-back is diagnostic only — wrapped so a CDP failure never fails the draw.

### Why the read-back

Silently dropping a requested style is the part that cost real time here. Accepting more key names fixes the specific case; reporting what actually landed makes *any* future mismatch visible in the response instead of only on the chart.

Happy to put it behind a flag or restrict it to the mismatch case if the extra `evaluate` per styled draw is a concern — it's one round-trip per shape.

### Tests

19 new tests in `tests/drawing_style.test.js`, wired into `test:unit` / `test:all`, following the existing `_deps` injection pattern from `sanitization.test.js`. They pin the shorthand mapping, `overrides` precedence, the colour normalizer, and the mismatch report — using the observed yellow/width-1/solid response as the negative case, so the regression itself is covered.

`npm run test:unit` → 152 existing + 19 new, all passing. `eslint` clean. **`test:e2e` not run** — it drives a live TradingView Desktop instance.


joebowen1208:
Verified against a live TradingView Desktop chart (not just the unit mocks), by calling the patched `drawShape` directly over the same CDP session so the change is exercised in isolation from a running server:

| Check | Result |
|---|---|
| `color` / `linewidth` / `linestyle` shorthand | `#FF80AB`, `2`, `2` — all applied |
| `overrides` as a JSON string (back-compat) | applied |
| `overrides` as an object (new) | applied |
| Precedence: explicit `overrides` beats shorthand | `#FFEB3B` won over `color: "#111111"` |
| Chart state after cleanup | 36 → 36 shapes, all probes removed |

Negative case, since a mismatch report that can never fire would be worth nothing:

```
style_requested  : {"linecolor":"#FF80AB","notARealProperty":42}
style_applied    : {"linecolor":"#FF80AB"}
style_not_applied: ["notARealProperty"]
```

The detector flags the key the chart ignored and does not flag the one that landed.

For reference, the same shorthand call on `main` returns `success: true` with `linecolor: rgba(255, 235, 59, 0.9809)` (TradingView's default yellow), width 1, solid — and says nothing about having dropped anything.

